PR #3, opened 2015-01-13
On "destroy()", don't flush the batch if it's empty.

bugflux:
This empty check could be added directly in the "flush()" method and avoid future regressions of this kind.
    However, by doing the change just here I'm going for:
        1) the least possible change impact;
        2) avoiding the unnecessary double-check on "add() -> full? -> flush()"

### DIFF
--- a/src/main/java/com/feedzai/commons/sql/abstraction/batch/AbstractBatch.java
+++ b/src/main/java/com/feedzai/commons/sql/abstraction/batch/AbstractBatch.java
@@ -144,7 +144,9 @@ public abstract class AbstractBatch implements Runnable {
             logger.debug("Interrupted while waiting.", e);
         }
 
-        flush();
+        if (batch != batchSize) {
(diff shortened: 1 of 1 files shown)


defer:
I'd go with the change to flush.

Semantically, flush() should be a no-op when there is nothing to flush. This is even more relevant when it is exposed as a public API _and_ it manipulates the internal state.

rfer:
Yes, flush() is also called from other methods, which will also benefit from it.

diogoeag:
I agree with @defer. I would go for something like "if(buffer.isEmpty()){ log and return. }

bugflux:
Agreed. Would you remove the check from the flush timeout handler, and have logs like:

    "Flush timeout occurred"
    "Batch empty, not flushing"

Or just log `"Flush timeout occurred"` when there's a timeout AND elements to flush?

PR #16, opened 2015-05-20
Fixes #15 by ignoring all system-generated columns in Oracle tables.

### LINKED ISSUE #15: updateEntity call fails when the table has system-generated columns
When using Oracle (tested with version 12c) `AbstractDatabaseEngine#updateEntity` fails when it tries to drop a system-generated column (name starts with "SYS_") from a table.

I'm not sure what causes these system columns to be added though.

The offending SQL query:

```
OracleEngine.dropColumn:519 ALTER TABLE "FDZ_APP_WORKFLOW_ELEMENT" DROP ( "SYS_STS348AI90XHRJWNMOT2K_PHJ#" )
```

And the exception stack trace:

```
com.feedzai.commons.sql.abstraction.engine.DatabaseEngineException: Error dropping column
at com.feedzai.commons.sql.abstraction.engine.impl.OracleEngine.dropColumn(OracleEngine.java:526)
at com.feedzai.commons.sql.abstraction.engine.AbstractDatabaseEngine.updateEntity(AbstractDatabaseEngine.java:517)
[...]
Caused by: java.sql.SQLException: ORA-12996: cannot drop system-generated virtual column
[...]
```

joaonuno:
When the internal engine state is populated, the columns whose name starts with
'SYS_' are not stored in the engine metadata about that table.

### DIFF
--- a/src/main/java/com/feedzai/commons/sql/abstraction/engine/impl/OracleEngine.java
+++ b/src/main/java/com/feedzai/commons/sql/abstraction/engine/impl/OracleEngine.java
@@ -820,11 +820,21 @@ public class OracleEngine extends AbstractDatabaseEngine {
             rsColumns = s.executeQuery(String.format("SELECT COLUMN_NAME, DATA_TYPE, DATA_PRECISION FROM ALL_TAB_COLS WHERE TABLE_NAME = '%s' AND OWNER = UPPER('%s')", name, properties.getProperty(PdbProperties.USERNAME)));
 
             while (rsColumns.next()) {
-                final String dataPrecision = rsColumns.getString("DATA_PRECISION");
+                String columnName = rsColumns.getString("COLUMN_NAME");
+                /*
+                Columns starting with SYS_ are Oracle system-generated columns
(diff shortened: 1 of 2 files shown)


src/main/java/com/feedzai/commons/sql/abstraction/engine/impl/OracleEngine.java
```diff
@@ -820,11 +820,21 @@ protected String getSchema() {
...
-                    metaMap.put(rsColumns.getString("COLUMN_NAME"), value);
+                String columnName = rsColumns.getString("COLUMN_NAME");
+                /*
+                Columns starting with SYS_ are Oracle system-generated columns
+                and PDB should not interfere with these columns. Not storing
+                these in the metaMap makes them invisible to PDB.
+
+                See: http://docs.oracle.com/cd/B19306_01/server.102/b14200/ap_keywd.htm
+                 */
+                if (!columnName.toUpperCase().startsWith("SYS_")) {
```
diogoeag: We should add a test to the OracleEngineSchemaTest right?

PR #18, opened 2015-09-04
fix PULSEDEV-9986 pdb: Underflow exception with oracle

leitaop:
Oracle JDBC driver throws underflow exeception whenver binding a value smaller or equal than 1e-131. Changed so that bound values smaller than 1e-31 are replaced by zero. Also, created automated unit tests for all scenarios with bind variables.

### DIFF
--- a/src/main/java/com/feedzai/commons/sql/abstraction/engine/impl/OracleEngine.java
+++ b/src/main/java/com/feedzai/commons/sql/abstraction/engine/impl/OracleEngine.java
@@ -72,6 +72,10 @@ public class OracleEngine extends AbstractDatabaseEngine {
      * Foreign key already exists
      */
     public static final String FOREIGN_ALREADY_EXISTS = "ORA-02275";
+    /**
+     *  Double instance for 0.0, so no Double instance is created whenever 0.0 is necessary.
+     */
+    private static Double ZERO = Double.valueOf(0.0);
 
     /**
      * Creates a new Oracle connection.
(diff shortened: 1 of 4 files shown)


src/test/resources/connections.properties
```diff
@@ -1,4 +1,10 @@
 h2.engine=com.feedzai.commons.sql.abstraction.engine.impl.H2Engine
 h2.jdbc=jdbc:h2:target/pdb
 h2.username=pdb
-h2.password=pdb
\ No newline at end of file
+h2.password=pdb
+
+# Example for oracle
+oracle.engine=com.feedzai.commons.sql.abstraction.engine.impl.OracleEngine
```
diogoeag: please remove this from default connections. custom connection.properties should be only local

src/test/java/com/feedzai/commons/sql/abstraction/engine/impl/oracle/OracleUnderflowTest.java
```diff
@@ -0,0 +1,197 @@
...
+ * Tests to ensure values less than 1.0e-131 are stored as 0. This is necessary because
+ * values less than 1.0e-131 provoke an underflow error in the Oracle JDBC driver when
+ * used in bind parameters.
+ *
+ * @author Paulo Leitao (paulo.leitao@feedzai.com)
+ *
+ * @since 2.1.4
+ */
+@RunWith(Parameterized.class)
+public class OracleUnderflowTest {
```
diogoeag: I would suggest to add this in the generic tests that run in all the databases. If the issue is fixed, then it should pass in all the databases.

src/main/java/com/feedzai/commons/sql/abstraction/engine/impl/OracleEngine.java
```diff
@@ -72,6 +72,10 @@
      * Foreign key already exists
      */
     public static final String FOREIGN_ALREADY_EXISTS = "ORA-02275";
+    /**
+     *  Double instance for 0.0, so no Double instance is created whenever 0.0 is necessary.
+     */
+    private static Double ZERO = Double.valueOf(0.0);
```
diogoeag: please add @since

PR #19, opened 2015-10-09
fix PULSEDEV-10540 pdb: Double serialization in event storage for blob fields

leitaop:
Changed handling of BLOB values so that byte arrays are not serialized. Up to now, BLOB values are always stored as java serialized values; with this change, BLOB values will have a prefix (encoder id) that indicates how the value is encoded followed by the actual value. Two encodings are supported, java serialization (the old encoding) plus null serialization (that from now on will be used for byte arrays).

In order to have backward compatibility, the java serializer encoding id is 0xac, which is a prefix used by the java serialization protocol (0xac is the first byte of STREAM_MAGIC). In addition, on this encoding, the encoding id is considered part of the encoded value.

Existing unit tests already cover scenarios for BLOB byte[] values and values of other types. Added only scenario for BLOB null values.


src/main/java/com/feedzai/commons/sql/abstraction/engine/BlobEncoder.java
```diff
@@ -0,0 +1,140 @@
...
+    }
+
+    /**
+     * Converts an object to a byte array representation in a BLOB value.
+     *
+     * @param val The object to convert.
+     * @return The byte array representation of the object.
+     * @throws IOException If the buffer is not enough to make the conversion.
+     */
+    public final synchronized byte[] encode(Object val) throws IOException {
```
bernardino: we should replace synchronized with a ReentrantLock and use it only in the `else` part since that is where we need it due to the reusable byte buffer

src/main/java/com/feedzai/commons/sql/abstraction/engine/BlobEncoder.java
```diff
@@ -0,0 +1,151 @@
...
+        if (val instanceof  byte[]) {
+            // Byte arrays can be saved with no serialization, just add encoder id
+            byte[] arrayValue = (byte[]) val;
+            byte[] encodedValue = new byte[arrayValue.length + 1];
+            encodedValue[0] = ID_NULL_ENCODER;
+            System.arraycopy(arrayValue, 0, encodedValue, 1, arrayValue.length);
+            return encodedValue;
+
+        } else {
+            // All other types are encoded with standard java serialization
```
rfer: Add trace logging indicating if enconding or not.

rfer:
Overall it looks fine to me

PR #21, opened 2015-10-16
Implements add column support

joaonuno:


joaonuno:
I am closing this PR because adding a column this way would cause inconsistencies to the table metadata.

PR #32, opened 2016-04-22
ref PULSEDEV-12907 pdb: Support complex fields as JSON/JSONB

leitaop:
-  Added DbColumnType.JSON, that is mapped either to a native JSON type whenever the engine supports it or to a CLOB otherwise.
-  Added method setParameter(final String name, final int index, final Object param, DbColumnType paramType) to DatabaseEngine, that allows specifying the intended type of a PreparedStatement parameter.  This is meant only for when it is not possible to figure this out from the Java type, which may happen now on String values: by default they represent String values (as before the change), but they may also represent JSON values.
 -  Fixed UnderflowTest unit test, that had hardcoded SQL on prepared statements that was not portable across DB engines.

### DIFF
--- a/src/main/java/com/feedzai/commons/sql/abstraction/ddl/DbColumnType.java
+++ b/src/main/java/com/feedzai/commons/sql/abstraction/ddl/DbColumnType.java
@@ -51,6 +51,11 @@ public enum DbColumnType {
      */
     CLOB,
     /**
+     * The json type
+     * @since 2.1.6
+     */
+    JSON,
+    /**
      * A type that is not mapped.
      */
     UNMAPPED;
--- a/src/main/java/com/feedzai/commons/sql/abstraction/ddl/DbColumnType.java
+++ b/src/main/java/com/feedzai/commons/sql/abstraction/ddl/DbColumnType.java
@@ -51,8 +51,9 @@ public enum DbColumnType {
      */
     CLOB,
     /**
-     * The json type
-     * @since 2.1.6
+     * The json type. This is not supported by all engines; engines not
+     * supporting it use the CLOB type.
+     * @since 2.1.5
      */
     JSON,
     /**
--- a/src/main/java/com/feedzai/commons/sql/abstraction/dml/result/PostgreSqlResultColumn.java
+++ b/src/main/java/com/feedzai/commons/sql/abstraction/dml/result/PostgreSqlResultColumn.java
@@ -39,11 +39,12 @@ public class PostgreSqlResultColumn extends ResultColumn {
      * Overrides default behaviour for JSON values, that are converted to strings.
      *
      * @param o The object in need of some kind of processing before being set.
-     * @return
+     * @return  The processed object.
+     * @since 2.1.5
      */
     @Override
     protected Object processObject(Object o) {
-        if (o != null  && o instanceof PGobject &&
+        if (o instanceof PGobject &&
                         ((PGobject)o).getType().equals("jsonb")) {
             return ((PGobject) o).getValue();
         }
--- a/src/main/java/com/feedzai/commons/sql/abstraction/dml/result/PostgreSqlResultColumn.java
+++ b/src/main/java/com/feedzai/commons/sql/abstraction/dml/result/PostgreSqlResultColumn.java
@@ -15,6 +15,8 @@
  */
 package com.feedzai.commons.sql.abstraction.dml.result;
 
+import org.postgresql.util.PGobject;
+
 /**
  * The PostgreSql column result implementation.
  *
@@ -31,5 +33,20 @@ public class PostgreSqlResultColumn extends ResultColumn {
      */
     public PostgreSqlResultColumn(final String name, final Object val) {
         super(name, val);
+    }
+
+    /**
+     * Overrides default behaviour for JSON values, that are converted to strings.
+     *
+     * @param o The object in need of some kind of processing before being set.
+     * @return
+     */
+    @Override
+    protected Object processObject(Object o) {
+        if (o != null  && o instanceof PGobject &&
+                        ((PGobject)o).getType().equals("jsonb")) {
+            return ((PGobject) o).getValue();
+        }
+        return super.processObject(o);
     }
 }
--- a/src/main/java/com/feedzai/commons/sql/abstraction/engine/AbstractDatabaseEngine.java
+++ b/src/main/java/com/feedzai/commons/sql/abstraction/engine/AbstractDatabaseEngine.java
@@ -1360,6 +1360,8 @@ public abstract class AbstractDatabaseEngine implements DatabaseEngine {
      * Maps the database type to {@link DbColumnType}. If there's no mapping a {@link DbColumnType#UNMAPPED} is returned.
      *
      * @param type The SQL type from {@link java.sql.Types}.
+     * @param typeName  The native database type name.  It provides additional information for
+     *                  derived classes to resolve types unmapped here.
      * @return The {@link DbColumnType}.
      */
     protected DbColumnType toPdbType(final int type, final String typeName) {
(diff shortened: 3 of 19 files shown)


src/main/java/com/feedzai/commons/sql/abstraction/dml/result/PostgreSqlResultColumn.java
```diff
@@ -32,4 +34,19 @@
...
         super(name, val);
     }
+
+    /**
+     * Overrides default behaviour for JSON values, that are converted to strings.
+     *
+     * @param o The object in need of some kind of processing before being set.
+     * @return
+     */
+    @Override
```
diogoeag: @since

src/main/java/com/feedzai/commons/sql/abstraction/dml/result/PostgreSqlResultColumn.java
```diff
@@ -32,4 +34,19 @@
     public PostgreSqlResultColumn(final String name, final Object val) {
         super(name, val);
     }
+
+    /**
+     * Overrides default behaviour for JSON values, that are converted to strings.
+     *
+     * @param o The object in need of some kind of processing before being set.
+     * @return
```
rfer: Missing javadoc

src/main/java/com/feedzai/commons/sql/abstraction/engine/AbstractDatabaseEngine.java
```diff
@@ -1361,7 +1362,7 @@ protected String getSchema() {
      * @param type The SQL type from {@link java.sql.Types}.
      * @return The {@link DbColumnType}.
      */
-    protected DbColumnType toPdbType(final int type) {
+    protected DbColumnType toPdbType(final int type, final String typeName) {
```
rfer: Missing typeName javadoc

src/main/java/com/feedzai/commons/sql/abstraction/dml/result/PostgreSqlResultColumn.java
```diff
@@ -32,4 +34,19 @@
...
+     */
+    @Override
+    protected Object processObject(Object o) {
+        if (o != null  && o instanceof PGobject &&
```
defer: null check is not needed, instanceof will return false if the left operand is null.

src/main/java/com/feedzai/commons/sql/abstraction/engine/DatabaseEngine.java
```diff
@@ -418,6 +418,20 @@ void createPreparedStatement(final String name, final String query, final int ti
     void setParameter(final String name, final int index, final Object param) throws DatabaseEngineException, ConnectionResetException;
 
     /**
+     * Sets the parameter on the specified index given its type. This is for situations where the java type of the parameter
```
rfer: An example of such situation would be nice.

src/main/java/com/feedzai/commons/sql/abstraction/engine/impl/DB2Engine.java
```diff
@@ -109,6 +109,7 @@ protected int entityToPreparedStatement(final DbEntity entity, final PreparedSta
                     /*
                      * CLOB and BLOB are handled the same way in DB2 since CLOB is not supported.
```
rfer: Include that JSON is not supported as well

src/main/java/com/feedzai/commons/sql/abstraction/engine/impl/DB2Translator.java
```diff
@@ -328,6 +328,7 @@ public String translate(DbColumn c) {
                 return format("VARCHAR(%s)", c.isSizeSet() ? c.getSize().toString() : properties.getProperty(VARCHAR_SIZE));
 
             /* DB2 does not support CLOB (or at least the Java driver is not implemented. */
```
rfer: Include that JSON is not supported as well

src/main/java/com/feedzai/commons/sql/abstraction/engine/impl/PostgreSqlEngine.java
```diff
@@ -126,6 +132,25 @@ protected int entityToPreparedStatement(final DbEntity entity, final PreparedSta
...
 
     @Override
+    public synchronized void setParameter(final String name, final int index, Object param, DbColumnType paramType) throws DatabaseEngineException, ConnectionResetException {
+        if (paramType == DbColumnType.JSON) {
+            param = getJSONValue((String)param);
+        }
+        super.setParameter(name, index, param);
+    }
+
+    private Object getJSONValue(String val) throws DatabaseEngineException {
```
rfer: Missing javadoc

src/main/java/com/feedzai/commons/sql/abstraction/engine/DatabaseEngine.java
```diff
@@ -418,6 +418,20 @@ void createPreparedStatement(final String name, final String query, final int ti
...
+     * alone is not enough to determine the correspondent database type.
```
defer: corresponding

src/main/java/com/feedzai/commons/sql/abstraction/engine/impl/PostgreSqlEngine.java
```diff
@@ -126,6 +132,25 @@ protected int entityToPreparedStatement(final DbEntity entity, final PreparedSta
...
+        try {
+            PGobject dataObject = new PGobject();
+            dataObject.setType("jsonb");
+            dataObject.setValue(val);
+            return dataObject;
+        } catch (SQLException ex) {
+            throw new DatabaseEngineException("Error while mapping variables to database", ex);
```
rfer: You have the value, could be worthwhile including it in the exception for more context.

src/main/java/com/feedzai/commons/sql/abstraction/ddl/DbColumnType.java
```diff
@@ -51,6 +51,11 @@
      */
     CLOB,
     /**
+     * The json type
```
rfer: I think it's important to note that this is not natively supported by all engines and is stored as a CLOB as last resort.

src/test/java/com/feedzai/commons/sql/abstraction/engine/impl/abs/JSonTest.java
```diff
@@ -0,0 +1,188 @@
...
+import static org.junit.Assert.assertThat;
+
+/**
+ * Tests for JSON columns.
+ *
+ * @author Paulo Leitao (paulo.leitao@feedzai.com)
+ * @since 2.1.6
+ */
+@RunWith(Parameterized.class)
+public class JSonTest {
```
defer: There should be a test for the case where json is invalid. I think that'll cause issues in native implementations but not in CLOBS but a test would clear up exactly what happens.

src/test/java/com/feedzai/commons/sql/abstraction/engine/impl/abs/JSonTest.java
```diff
@@ -0,0 +1,188 @@
...
+     * Checks that the test table has a single entry and with the values as expected.
+     */
+    private void checkInsertedValue(String jsonPart1, String jsonPart2) throws DatabaseEngineException {
+        Expression query = select(all()).from(table(TEST_TABLE));
+        List<Map<String, ResultColumn>> results = dbEngine.query(query);
+        assertEquals("One value inserted", 1, results.size());
+        Map<String, ResultColumn> firstRow = results.get(0);
+        assertNotNull("Inserted row is not null", firstRow);
+        assertThat("JSon value is as expected", firstRow.get(JSON_COL).toString(), anyOf(
+                is("{" + jsonPart1 + ", " + jsonPart2 + "}"),
```
stanch: This might fail on an engine that trims spaces when returning JSON. I would just trim them.

defer:
lgtm.

PR #33, opened 2016-04-27
Rolls back the transaction on batch flush failure

joaonuno:
Before this fix, when flushing the batch, if the addBatch call failed the transaction
would not be rolled back because the roll back was only being done in the inner try-catch.
This PR just moves the addBatch loop to inside the inner try-catch. It also adds a regression
unit-test.

### DIFF
--- a/src/main/java/com/feedzai/commons/sql/abstraction/batch/AbstractBatch.java
+++ b/src/main/java/com/feedzai/commons/sql/abstraction/batch/AbstractBatch.java
@@ -25,7 +25,6 @@ import org.slf4j.LoggerFactory;
 import org.slf4j.Marker;
 import org.slf4j.MarkerFactory;
 
-import java.util.ArrayList;
 import java.util.LinkedList;
 import java.util.List;
 import java.util.concurrent.Executors;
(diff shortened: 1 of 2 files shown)


src/main/java/com/feedzai/commons/sql/abstraction/batch/AbstractBatch.java
```diff
@@ -58,11 +57,23 @@
...
      *
      * @since 2.1.4
      */
-    private final Lock flushLock = new ReentrantLock();
+    private final Lock bufferLock = new ReentrantLock();
+
+    /**
+     * Lock used for concurrent access to the flush transaction.
+     * <p>
+     * We are not using the {@link #bufferLock} because the
```
diogoeag: we are not using?

joaonuno: I'm explaining why we are not using the bufferLock for locking the actual transaction flush

diogoeag: SSorry, because of the highlight I was reading the flushTransactionLock javadoc as it would be bufferLock

PR #36, opened 2016-05-04
Loss of precision when using ResultColumn.toLong()

leitaop:
Removed usage of the intermediate double when the stored value has no decimal fields.  When it has decimal fields, the intermediate double can be used as before because it means the value was not stored as a long.


rfer:
LGTM.

PR #39, opened 2016-06-30
Fix DDL change detection for PostgreSqlEngine

bugflux:
PostgreSqlEngine always fetches metadata from the "public" schema when trying to detect table changes, regardless of the schema configured for the Database Engine. This hotfix corrects this behaviour.


orium:
@rfer can you expedite this PR?

PR #44, opened 2016-09-19
ref PULSEDEV-14916 pdb: Added support for synchronous/serial execution of AbstractBacth#flush

diogoeag:
- When executing #flush(true) the batch will wait for other batches executing.

### DIFF
--- a/src/main/java/com/feedzai/commons/sql/abstraction/batch/AbstractBatch.java
+++ b/src/main/java/com/feedzai/commons/sql/abstraction/batch/AbstractBatch.java
@@ -203,27 +203,7 @@ public abstract class AbstractBatch implements Runnable {
      * @implSpec Same as {@link #flush(boolean)} with {@link false}.
      */
     public void flush() {
-        flush(false);
-    }
-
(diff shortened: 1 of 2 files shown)


src/main/java/com/feedzai/commons/sql/abstraction/batch/AbstractBatch.java
```diff
@@ -199,10 +199,32 @@ public void add(final String entityName, final EntityEntry ee) throws DatabaseEn
...
+     * @param sync {@literal true} if it should wait for other {@link #flush()} operations.
+     * @implSpec It is possible that two threads might execute in competing order and sync execution acquires the flushTransactionLock before a non synchronous one leading
+     * to a non serial execution.
+     * @since 2.1.6
+     */
+    public void flush(boolean sync) {
         List<BatchEntry> temp;
 
+        if (sync) {
+            flushTransactionLock.lock();
```
rfer: Ideally this lock should be followed by a try finally, in case something goes wrong.

Issue #45, opened 2016-09-29
NaN values will cause insertions to fail in Oracle 12


jprafael:
The `OracleTranslator` is currently converting `DbColumnType.DOUBLE` to oracle's `DOUBLE PRECISION`. According to [oracle's documentation](https://docs.oracle.com/cd/E11882_01/timesten.112/e21642/types.htm) we should be using `BINARY_DOUBLE` instead since that is what conforms to the IEEE standard - also what is supported by Java and many other databases.

As a consequence of not doing so we fall into the limitations of `DOUBLE PRECISION`
 * No support for NaN, -Inf or Inf.
 * Different floating point handling which caused https://github.com/feedzai/pdb/pull/18

Changing the column format and a little bit more code in `OracleEngine.entityToPreparedStatement` fixes this.

This however has the downside of being not backwards compatible and potentially requiring changes in multiple client's DBs. 

Let me know how you want to handle that.

PR #46, opened 2016-11-07
ref PULSEDEV-15386 pdb: Changed Double data type used for oracle columns

Mikah89:
To allow the insertion of values such as NaN and infinity the double column data type for oracle was changed to BINARY_DOUBLE (from DOUBLE PRECISION).

This addresses the bug reported in #45 .
The issue description mentioned something about changing code in the `OracleEngine.entityToPreparedStatement` but I didn't see the need for that as far as my understanding goes. Please double check that and warn me if there is other things to do here.

The test is pending for some machines to be up (already sent email) to check if it breaks any test. I should create a new UT afterwards.

### DIFF
--- a/src/main/java/com/feedzai/commons/sql/abstraction/engine/impl/OracleEngine.java
+++ b/src/main/java/com/feedzai/commons/sql/abstraction/engine/impl/OracleEngine.java
@@ -924,6 +924,7 @@ public class OracleEngine extends AbstractDatabaseEngine {
                 return DbColumnType.BOOLEAN;
             case "FLOAT":
             case "FLOAT126":
+            case "BINARY_DOUBLE":
                 return DbColumnType.DOUBLE;
             case "LONG":
             case "NUMBER19":
--- a/src/main/java/com/feedzai/commons/sql/abstraction/engine/impl/OracleEngine.java
+++ b/src/main/java/com/feedzai/commons/sql/abstraction/engine/impl/OracleEngine.java
@@ -23,6 +23,7 @@ import com.feedzai.commons.sql.abstraction.engine.*;
 import com.feedzai.commons.sql.abstraction.engine.configuration.PdbProperties;
 import com.feedzai.commons.sql.abstraction.engine.handler.OperationFault;
 import com.feedzai.commons.sql.abstraction.entry.EntityEntry;
+import com.feedzai.commons.sql.abstraction.util.PreparedStatementCapsule;
 import com.google.common.base.Optional;
 import com.google.common.base.Preconditions;
 import com.google.common.collect.ImmutableList;
@@ -147,6 +148,9 @@ public class OracleEngine extends AbstractDatabaseEngine {
                         }
 
                         break;
+                    case DOUBLE:
+                        setParameterValues(ps, i, val);
+                        break;
                     default:
                         ps.setObject(i, ensureNoUnderflow(val));
                 }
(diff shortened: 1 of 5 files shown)


Mikah89:
Just found some more changes to do after running the existing UTs :)

src/test/java/com/feedzai/commons/sql/abstraction/engine/impl/oracle/OracleEngineSchemaTest.java
```diff
@@ -100,4 +106,51 @@ public void testSystemGeneratedColumns() throws Exception {
...
         }
     }
+
+    @Test
+    public void testInsertNan() throws Exception {
+        testInsertSpecialValues("NaN");
+    }
+
+    @Test
+    public void testInsertInfinity() throws Exception {
```
henriquevcosta: Please document tests, even if evident.

src/test/java/com/feedzai/commons/sql/abstraction/engine/impl/oracle/OracleEngineSchemaTest.java
```diff
@@ -17,19 +17,25 @@
```
leitaop: I think all these tests could all be in AbstractEngineSchemaTest, since they are not Oracle-specific.

src/test/java/com/feedzai/commons/sql/abstraction/engine/impl/oracle/OracleEngineSchemaTest.java
```diff
@@ -100,4 +106,68 @@ public void testSystemGeneratedColumns() throws Exception {
...
+     */
+    @Test
+    public void testInsertInfinity() throws Exception {
+        testInsertSpecialValues("Infinity");
+    }
+
+    /**
+     * The 'randomString' is not a special value for the BINARY_DOUBLE type so it should throw an error.
+     */
+    @Test
```
leitaop: You can simplify the test with 
@Test(expected=Exception.class)

src/test/java/com/feedzai/commons/sql/abstraction/engine/impl/oracle/OracleEngineSchemaTest.java
```diff
@@ -100,4 +106,68 @@ public void testSystemGeneratedColumns() throws Exception {
...
+                    .build();
+            engine.addEntity(entity);
+
+            final EntityEntry entry = entry()
+                    .set("id", 1)
+                    .set("DBL", columnValue)
+                    .build();
+
+            engine.persist(entity.getName(), entry);
+            final List<Map<String, ResultColumn>> dbl = engine.query(select(column("DBL")).from(table(entity.getName())));
```
leitaop: This is only one way of inserting the value, you can see in LongOverflowTest examples of other ways to do it that are not covered in this test.

leitaop:
LGTM

src/main/java/com/feedzai/commons/sql/abstraction/engine/impl/OracleEngine.java
```diff
@@ -160,6 +164,38 @@ protected int entityToPreparedStatement(final DbEntity entity, final PreparedSta
...
+                 * Only Double typed values will execute here.
+                 * They may be:
+                 *  - Regular double number
+                 *  - Double.NaN
+                 *  - Double.{POSITIVE|NEGATIVE}_INFINITY
+                 *
+                 *  In these cases we use the setBinaryDouble specific to the oracle prepared statement.
+                 */
+                final OraclePreparedStatement orclPs = (OraclePreparedStatement) ps;
+                orclPs.setBinaryDouble(index, (Double) ensureNoUnderflow(param));
```
leitaop: Does this work in the old native type for double ?  If not, maybe we can use setBinaryDouble() only if the value is NaN or Infinity, and use setObject().

leitaop:
Checked that double operations still work on tables created with the old type assigned to double (DOUBLE_PRECISION).

Issue #47, opened 2016-11-17
pdb.allow_column_drop defaults to true and is very risky


diogoeag:
When a connection.updateEntity is executed, if the new version of the entity has less colums, pdb will drop those columns by default.

This is a risky and error prone behavior

PR #49, opened 2016-12-16
ref PULSEDEV-15861: Reverted Allowed column drop default value.

Mikah89:
The change in PR #47 caused some problems on components using the pdb. So the value was reverted to true again. Any other desired value for this property should be configured to override this setting.

Rewrote the unit test to check that without configuration the default is to drop columns.


defer:
Can you summarize what problems it caused for future reference?

PR #50, opened 2016-12-29
ref PULSEDEV-16501 pdb: Improve the performance of writing to Oracle …

nmldiegues:
…CLOB columns

The usage of PreparedStatement#setClob in Oracle 11g was yielding overheads of 3x
to 1000x (depending on the batch size) when compared to using plain
PreparedStatement#setObject for columns backed by VARCHAR2.

With this change we downcast the PreparedStatement to OraclePreparedStatement and
use its setStringForClob method that yields equivalent performance to the VARCHAR2
case for Strings whose length is 32KB or less.

### DIFF
--- a/src/main/java/com/feedzai/commons/sql/abstraction/engine/impl/OracleEngine.java
+++ b/src/main/java/com/feedzai/commons/sql/abstraction/engine/impl/OracleEngine.java
@@ -30,7 +30,6 @@ import com.google.common.collect.ImmutableList;
 import oracle.jdbc.OraclePreparedStatement;
 import oracle.jdbc.OracleTypes;
 
-import java.io.StringReader;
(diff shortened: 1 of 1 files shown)


diogoeag:
Seems fine to me!

nmldiegues:
I tested this manually by using the migrator tool for event storage 15.1 to 16.1 which ends up calling this code. Then checked to see that the contents were as expected.

I also ran the UTs suite against an Oracle instance, but I have yet to confirm whether there is any test that covers this code. Will do that before merging the pull request.

nmldiegues:
I have double checked that there are various UTs that use CLOB/JSON column type with String objects, and thus exercise the line of code that was changed. So we are good to go.

PR #52, opened 2017-02-24
Creates a new factory method in DatabaseFactory

henriquevcosta:
that takes a PdbProperties object instead of a
Properties.


henriquevcosta:
Ignore this.

PR #57, opened 2017-04-13
Add indexes when a new column is added to an entity

donbonifacio:
When a column is being added to a table that already exists, it's indexes
were being left out. This patch fixes that by adding the indexes that
consider the new column(s).

A new `addIndexes` method was added to `AbstractDatabaseEngine` that
creates specific indexes. I didn't want to change the existing method
not to break the interface.

Missing on this patch:
* Wasn't able to test this, missing dependencies and couldn't find
  instructions
* Would like a unit test for this, but don't have the env and wouldn't
  know how to achieve it for this scenario

Related possible future work:
* Implement drop indexes
* When we remove columns, we should also remove related indexes
  (I guess it would throw an exception otherwise).
* Use the same logic for FKs. ATM we're droping/adding them all when
  adding or removing new fields. That could be very slow on large tables

### DIFF
--- a/src/main/java/com/feedzai/commons/sql/abstraction/engine/AbstractDatabaseEngine.java
+++ b/src/main/java/com/feedzai/commons/sql/abstraction/engine/AbstractDatabaseEngine.java
@@ -31,6 +31,7 @@ import com.feedzai.commons.sql.abstraction.util.InitiallyReusableByteArrayOutput
 import com.feedzai.commons.sql.abstraction.util.PreparedStatementCapsule;
 import com.google.inject.Inject;
 import com.google.inject.Injector;
+import java.util.stream.Collectors;
 import org.apache.commons.lang.StringUtils;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
@@ -519,6 +520,7 @@ public abstract class AbstractDatabaseEngine implements DatabaseEngine {
                 if (!toRemove.isEmpty()) {
                     if (properties.allowColumnDrop()) {
                         dropColumn(entity, toRemove.toArray(new String[toRemove.size()]));
+                        // TODO: drop indexes
                     } else {
                         logger.warn("Need to remove {} columns to update {} entity, but property allowColumnDrop is set to false.", StringUtils.join(toRemove, ","), entity.getName());
                     }
@@ -531,7 +533,9 @@ public abstract class AbstractDatabaseEngine implements DatabaseEngine {
                     }
                 }
                 if (!columns.isEmpty()) {
-                    addColumn(entity, columns.toArray(new DbColumn[columns.size()]));
+                    final DbColumn[] dbColumns = columns.toArray(new DbColumn[columns.size()]);
+                    addColumn(entity, dbColumns);
+                    addIndexes(entity, dbColumns);
                 }
                 addFks(entity);
             }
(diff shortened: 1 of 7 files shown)


src/main/java/com/feedzai/commons/sql/abstraction/engine/AbstractDatabaseEngine.java
```diff
@@ -519,6 +520,7 @@ public synchronized void updateEntity(DbEntity entity) throws DatabaseEngineExce
                 if (!toRemove.isEmpty()) {
                     if (properties.allowColumnDrop()) {
                         dropColumn(entity, toRemove.toArray(new String[toRemove.size()]));
+                        // TODO: drop indexes
```
diogoeag: forgot something?

src/main/java/com/feedzai/commons/sql/abstraction/engine/AbstractDatabaseEngine.java
```diff
@@ -546,6 +550,33 @@ public synchronized void updateEntity(DbEntity entity) throws DatabaseEngineExce
...
     }
 
+    /**
+     * Given a {@link DbEntity}, returns the defined indexes that include the given {@link DbColumn}.
+     *
+     * @param entity The owner {@link DbEntity}.
+     * @param column The {@link DbColumn} to search in indexes.
+     * @return The list of {@link DbIndex}s.
+     */
+    private List<DbIndex> getIndexesForColumn(final DbEntity entity, final DbColumn column) {
```
diogoeag: pleease add @since

src/main/java/com/feedzai/commons/sql/abstraction/engine/AbstractDatabaseEngine.java
donbonifacio: Not forgotten. PDB doesn't have yet the functionality to drop indexes, it does not do it anywhere, but should do it here. It's more of a reminder.

donbonifacio:
Hey @diogoeag , like I mentioned on the summary, I was unable to compile the project and run tests because of dependencies. And I didn't find any instructions on it. Also, I didn't find unit tests on `addIndexes` that I could build upon.

Would really like to do it, but will need more time to get this going. :)

nmldiegues:
Usually people fall into this issue, which is in the README

> Compiling PDB

> In order to compile PDB you will need to have the Oracle Driver JAR in your local repository. The current > version assumes Oracle Driver version 11.2.0.2.0. Please download the driver [here] and run the following to install the driver in your local maven repository. [here]:http://www.oracle.com/technetwork/database/enterprise-edition/jdbc-112010-090769.html

> mvn install:install-file -DgroupId=com.oracle -DartifactId=ojdbc6 \
> -Dversion=11.2.0.2.0 -Dpackaging=jar -Dfile=ojdbc6-11.2.0.2.0.jar


Could that be your problem?

donbonifacio:
@nmldiegues definitely it is. :)

How about running the tests? I assume that the tests are against all supported databases, and that would be tricky to setup.

May I suggest a Travis CI setup for this repo that setups all DBs, compiles, runs all the tests, etc, for each PR? I don't mind setting this up, if i have proper access.

There is also related future work here, not only on dropping indexes, but also doing the same for the foreign keys. At this moment they are all dropped and added again when we have new fields. But that for big databases will incur in a lot of unnecessary work.

donbonifacio:
Hey,

Just updated the commit with `@since` annotations. Was also able to compile successfully.

PR #58, opened 2017-04-22
ref PULSEDEV-18139 pdb: Fix incorrect shutdown in AbstractBatch

henriquevcosta:
This change fixes a small concurrency issue that was occurring in AbstractBatch
when the batch scheduler was scheduling a run while the destroy
method had just been called, leading to an unexpected long wait for the
destroy method to return.
Since those two methods were `synchronized`
the following could happen:
   - Thread A entered `destroy` thus acquiring the implicit lock
   - The scheduler period would expire and a new task would be
   launched but since the `run` method was synchronized the task's
   thread (B)  would not proceed
   - Thread A would now call `shutdownNow` on the ExecutorService and
   await termination.
   - Since Thread B was waiting on the implicit lock to enter `run` the
     `destroy` code would wait until the timeout expired before
     returning.

There was no reason to keep the methods synchronized as synchronization was
already being performed at a different level.

### DIFF
--- a/src/main/java/com/feedzai/commons/sql/abstraction/batch/AbstractBatch.java
+++ b/src/main/java/com/feedzai/commons/sql/abstraction/batch/AbstractBatch.java
@@ -101,7 +101,7 @@ public abstract class AbstractBatch implements Runnable {
     /**
      * Timestamp of the last flush.
      */
-    protected long lastFlush;
+    protected volatile long lastFlush;
     /**
      * EntityEntry buffer.
      */
--- a/src/main/java/com/feedzai/commons/sql/abstraction/batch/AbstractBatch.java
+++ b/src/main/java/com/feedzai/commons/sql/abstraction/batch/AbstractBatch.java
@@ -153,13 +153,21 @@ public abstract class AbstractBatch implements Runnable {
      */
     public void destroy() {
         logger.trace("{} - Destroy called on Batch", name);
-        scheduler.shutdownNow();
+        scheduler.shutdown();
 
         try {
             if (!scheduler.awaitTermination(maxAwaitTimeShutdown, TimeUnit.MILLISECONDS)) {
-                logger.warn("Could not terminate batch within {}", DurationFormatUtils.formatDurationWords(maxAwaitTimeShutdown, true, true));
+                logger.warn(
+                        "Could not terminate batch within {}. Forcing shutdown.",
+                        DurationFormatUtils.formatDurationWords(
+                                maxAwaitTimeShutdown,
+                                true,
+                                true
+                        )
+                );
+                scheduler.shutdownNow();
             }
-        } catch (InterruptedException e) {
+        } catch (final InterruptedException e) {
             Thread.currentThread().interrupt();
             logger.debug("Interrupted while waiting.", e);
         }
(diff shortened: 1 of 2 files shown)


donbonifacio:
Hey @henriquevcosta ,

Can you please expand on:

> There was no reason to keep the methods synchronized as synchronization was
already being performed at a different level.

I was looking at this and with this patch we can have a scheduled `flush` and a `destroy` at the same time. I'm wondering if the flush lock is enough or if a `shutdownNow` in the middle of a `flush` can leave things messy.

Thanks. :)

henriquevcosta:
Firstly, we could already have a flush at the same time since it is public but isn't synchronized. 
However, you are correct that now a running flush is more likely to get an interrupt because of the shutdownNow, but I don't think that synchronization is the way to go since it would lead to the same problem as before. We can, however, change this to a `shutdown` instead of a `shutdownNow` to attempt to minimize the likelihood. WDYT?

donbonifacio:
I'm assuming that after a `shutdown` the scheduler won't run again. So if we have a `flush` from the scheduler running, the immediate `awaitTermination ` will will wait for it to finish, and if it timesout then we can brutally `shutdownNow` it.

Am I properly interpreting your suggestion? It sounds good to me.

src/main/java/com/feedzai/commons/sql/abstraction/batch/AbstractBatch.java
```diff
@@ -302,7 +302,7 @@ public void flush(boolean sync) {
     }
 
     @Override
-    public synchronized void run() {
+    public void run() {
         if (System.currentTimeMillis() - lastFlush >= batchTimeout) {
```
nmldiegues: This lastFlush variable should be made volatile given the removal of the synchronized. Otherwise there's no guarantee that this thread will see it updated (if the flush call is performed from an external thread via the public api of the batch, or if the executor had more threads --- which is not the case currently).

src/test/java/com/feedzai/commons/sql/abstraction/engine/impl/abs/BatchUpdateTest.java
```diff
@@ -287,6 +287,39 @@ void beginTransaction(Invocation inv) throws DatabaseEngineRuntimeException {
...
+                .addColumn("COL2", BOOLEAN)
+                .addColumn("COL3", DOUBLE)
+                .addColumn("COL4", LONG)
+                .addColumn("COL5", STRING).build();
+
+        engine.addEntity(entity);
+
+
+        for (int i = 0; i < 40; i++) {
+            final MockedBatch batch = MockedBatch.create(engine, "test", 5, 10L, 50000);
```
nmldiegues: Add a comment expliciting that the "wait" timeout defined here is larger than the test timeout. Thus, if the regression was ever hit, it would generate a test timeout.

src/main/java/com/feedzai/commons/sql/abstraction/batch/AbstractBatch.java
henriquevcosta: Nice catch, will make it volatile

nmldiegues:
I'll let @donbonifacio say whether he is fully happy with this, since the conversation seems to still be on going?

PR #62, opened 2017-07-19
Add Support for Failure Listener when using pdb batches

Mikah89:
When executing batches if there were any errors while persisting data the behavior to deal with these errors had to be implemented in specific batch classes. Namely in Pulse we had a specific batch to deal with write failures which did not respect the pdp API.

Because of the above it was difficult for people to override the behavior implemented by Pulse. To fix this we implemented a default error handling function which receives a custom listener (from the client side). With this Pulse now respects the pdb API to create batches and thus it is easier to override pulse behavior when needed to.


src/main/java/com/feedzai/commons/sql/abstraction/batch/AbstractBatch.java
```diff
@@ -305,8 +338,22 @@ public void flush(boolean sync) {
...
+        if (failureListener.isPresent()) {
+            final FailureListener listener = this.failureListener.get();
+            final Set<Map<String, Serializable>> failedEvents = new HashSet<>();
+            for (final BatchEntry entry : entries) {
+                final Map<String, Object> eventMap = entry.getEntityEntry().getMap();
+                final Map<String, Serializable> event = Maps.newHashMapWithExpectedSize(eventMap.size());
+
+                // do not change this to collect(Collectors.toMap()) since it throws an NPE when the value is null
+                eventMap.keySet().
+                        forEach(name -> event.put(name, (Serializable) eventMap.get(name)));
```
marcobjorge: `event.putAll(eventMap);`?

Mikah89: yeh, this was basically reused from the failurebatch used in Pulse. But it might be a good idea for readability sake.

src/main/java/com/feedzai/commons/sql/abstraction/batch/AbstractBatch.java
```diff
@@ -305,8 +338,22 @@ public void flush(boolean sync) {
...
      */
-    public abstract void onFlushFailure(BatchEntry[] entries); {
-        // NO-OP.
+    public void onFlushFailure(BatchEntry[] entries) {
+        if (failureListener.isPresent()) {
+            final FailureListener listener = this.failureListener.get();
+            final Set<Map<String, Serializable>> failedEvents = new HashSet<>();
+            for (final BatchEntry entry : entries) {
+                final Map<String, Object> eventMap = entry.getEntityEntry().getMap();
+                final Map<String, Serializable> event = Maps.newHashMapWithExpectedSize(eventMap.size());
```
bernardino: `event` should probably be a LinkedHashMap just like `eventMap` is

src/main/java/com/feedzai/commons/sql/abstraction/FailureListener.java
```diff
@@ -0,0 +1,40 @@
...
+public interface FailureListener {
+
+    /**
+     * Callback indicating that one or more rows have failed to be persisted.
+     * <p>
+     * Each set entry represent a row in which the map keys are the column name.
+     *
+     * @param rowsFailed Set with the row or rows that failed to be persisted.
+     */
+    void onFailure(Set<Map<String, Serializable>> rowsFailed);
```
jprafael: Shouldn't this be a list instead of a set? batch inserts might be order dependent in case the tables have something like an auto-increment column

src/main/java/com/feedzai/commons/sql/abstraction/batch/AbstractBatch.java
```diff
@@ -110,24 +117,49 @@
...
      *
      * @param de                   The database engine.
      * @param name                 The batch name (null or empty names are allowed, falling back to "Anonymous Batch").
      * @param batchSize            The batch size.
      * @param batchTimeout         The batch timeout.
      * @param maxAwaitTimeShutdown The maximum await time for the batch to shutdown.
+     * @param failureListener      The listener that will be invoked whenever some batch operation fail to persist.
      */
-    protected AbstractBatch(final DatabaseEngine de, String name, final int batchSize, final long batchTimeout, final long maxAwaitTimeShutdown) {
+    protected AbstractBatch(
```
jprafael: Its about time this class its constructed with a builder

src/main/java/com/feedzai/commons/sql/abstraction/batch/AbstractBatch.java
```diff
@@ -305,8 +338,22 @@ public void flush(boolean sync) {
      *
      * @param entries The entries that are pending to be persisted.
      */
-    public abstract void onFlushFailure(BatchEntry[] entries); {
-        // NO-OP.
+    public void onFlushFailure(BatchEntry[] entries) {
+        if (failureListener.isPresent()) {
```
jprafael: use early exits instead of nesting:

```
if (!failureListener.isPresent()) {
   return;
}
```

src/main/java/com/feedzai/commons/sql/abstraction/batch/AbstractBatch.java
jprafael: Why is this method copying the list of entries and each entry in the first place? Why not pass the `BatchEntry[]` directly?

jprafael: This will fail if the original value is not `Serializable`. If this is not possible by construction then change the `entry.getEntityEntry().getMap()` to return a `Map<String, Serializable>` instead.

src/main/java/com/feedzai/commons/sql/abstraction/FailureListener.java
Mikah89: This was passed as is of the previous implementation. As well as most of the oher comments made on this review. This to say, there is no specific reason for the majority of the questions you made. This code basically was a shift from what we had on pulse to PDB in order for people to be able to override behavior.

I'll address them as improvements as soon as everyone expected to comment has commented.

Mikah89: Which probably will mean more tests also

src/main/java/com/feedzai/commons/sql/abstraction/engine/DatabaseEngine.java
```diff
@@ -244,6 +246,23 @@
...
+     * fails to persist data.
+     * <p/>
+     * Please be sure to call {@link AbstractBatch#destroy() } before closing the session with the database.
+     *
+     * @param batchSize       The batch size.
+     * @param batchTimeout    If inserts do not occur after the specified time, a flush will be performed.
+     * @param batchName       The batch name.
+     * @param failureListener Failure listener to execute custom behavior when the batch fails to persist.
+     * @return The batch.
+     */
```
luiscleto: missing @ since for the new interface method

src/main/java/com/feedzai/commons/sql/abstraction/batch/DefaultBatch.java
```diff
@@ -65,4 +77,24 @@ public static DefaultBatch create(final DatabaseEngine de, final String name, fi
...
+     * <p>Starts the timertask.</p>
+     *
+     * @param de                   The database engine.
+     * @param name                 The batch name.
+     * @param batchSize            The batch size.
+     * @param batchTimeout         The batch timeout.
+     * @param maxAwaitTimeShutdown The maximum await time for the batch to shutdown.
+     * @param listener             The listener that will be invoked when batch fails to persist at least one data row.
+     * @return The Batch.
+     */
```
luiscleto: @ since

src/main/java/com/feedzai/commons/sql/abstraction/batch/AbstractBatch.java
smmf: Can't be done as simply as @marcobjorge is suggesting, because the generic types of the maps are not exactly the same. eventMap's values are Object and event's values are declared as Serializable.

src/main/java/com/feedzai/commons/sql/abstraction/batch/AbstractBatch.java
```diff
@@ -305,8 +338,22 @@ public void flush(boolean sync) {
...
+            final FailureListener listener = this.failureListener.get();
```
smmf: I suggest to move this declaration down, just before it is necessary.

src/main/java/com/feedzai/commons/sql/abstraction/batch/AbstractBatch.java
smmf: Actually, this "Serializable" thing came from when the code was moved here from Pulse. There is nothing in PDB that requires this to be a Serializable. Please keep it as Object. Also, I've checked the code on Pulse, and it is possible to use this as Object. So, this can be simplified a lot here.

src/main/java/com/feedzai/commons/sql/abstraction/batch/DefaultBatch.java
```diff
@@ -32,19 +35,28 @@
...
+     * Creates a new instance of {@link DefaultBatch}.
+     *
+     * @param de                   The database engine reference.
+     * @param name                 The batch name.
+     * @param batchSize            The batch size.
+     * @param batchTimeout         The timeout.
+     * @param maxAwaitTimeShutdown The maximum await time for the batch to shutdown.
+     * @param listener             The listener that will be invoked when batch fails to persist at least one data row.
+     */
+    protected DefaultBatch(final DatabaseEngine de, final String name, final int batchSize, final long batchTimeout, final long maxAwaitTimeShutdown, final Optional<FailureListener> listener) {
```
smmf: Same as before. Don't take Optional as parameter, just the listener.

src/main/java/com/feedzai/commons/sql/abstraction/batch/DefaultBatch.java
```diff
@@ -65,4 +77,24 @@ public static DefaultBatch create(final DatabaseEngine de, final String name, fi
...
+     * @param de                   The database engine.
+     * @param name                 The batch name.
+     * @param batchSize            The batch size.
+     * @param batchTimeout         The batch timeout.
+     * @param maxAwaitTimeShutdown The maximum await time for the batch to shutdown.
+     * @param listener             The listener that will be invoked when batch fails to persist at least one data row.
+     * @return The Batch.
+     */
+    public static DefaultBatch create(final DatabaseEngine de, final String name, final int batchSize, final long batchTimeout,
+                                      final long maxAwaitTimeShutdown, final Optional<FailureListener> listener) {
```
smmf: take Optional out.

PR #63, opened 2017-08-28
Have DatabaseEngine extend AutoCloseable to enable try-with-resources.

gcandal:


src/main/java/com/feedzai/commons/sql/abstraction/engine/DatabaseEngine.java
```diff
@@ -40,7 +40,7 @@
  * @author Rui Vilao (rui.vilao@feedzai.com)
  * @since 2.0.0
  */
-public interface DatabaseEngine {
+public interface DatabaseEngine extends AutoCloseable {
```
joaonuno: Add the explicit @Override to method close

jmf-tls:
(not sure if)
This is a duplicate of #4 which is more complete, by the way (ResultIterator can also be closed).
The only thing requested in that one to be approved/landed is to have tests to make sure resources are closed after leaving a try-with-resources...

PR #4, opened 2015-01-13
Implement the java.lang.AutoCloseable interface where relevant

bugflux:
... to let clients use the Java 7+ "try-with-resources" feature.

PR #65, opened 2017-09-19
Clear batch data when rolling back a transaction to ensure ACID properties

dspereira004:
Currently, a transaction rollback executes a rollback statement on the
database. However PDB might still have prepared statements with batch data that,
unless cleared, can still be saved on the database, thereby violating the ACID
properties. This commit ensures that the prepared statements are cleared when
rolling back a transaction.

### DIFF
--- a/src/main/java/com/feedzai/commons/sql/abstraction/batch/AbstractBatch.java
+++ b/src/main/java/com/feedzai/commons/sql/abstraction/batch/AbstractBatch.java
@@ -185,7 +185,17 @@ public abstract class AbstractBatch implements Runnable {
      * Starts the timer task.
      */
     protected void start() {
-        scheduler.scheduleAtFixedRate(this, 0, batchTimeout + salt, TimeUnit.MILLISECONDS);
+        // if a periodic execution throws an exception, future executions are suspended,
+        // this task wraps the call in a try-catch block to prevent that. Errors are still propagated.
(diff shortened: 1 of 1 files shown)


src/main/java/com/feedzai/commons/sql/abstraction/engine/AbstractDatabaseEngine.java
```diff
@@ -692,6 +692,10 @@ public synchronized void rollback() throws DatabaseEngineRuntimeException {
          * Reconnect on this method does not make sense since a new connection will have nothing to rollback.
          */
         try {
+            // While rolling back, the entities can still have batch data which needs to be cleared.
+            for (final MappedEntity mappedEntity : entities.values()) {
+                mappedEntity.getInsert().clearBatch();
```
luiscleto: Shouldn't this also be done for 'getInsertWithAutoInc()' and 'getInsertReturning()' ?

dspereira004: Yes, you're right. I've updated the pull request.

luiscleto:
LGTM. Should probably have some unit tests though.
I'm torn between the trade-off of requesting tests and having another scenario like https://github.com/feedzai/pdb/pull/4 and https://github.com/feedzai/pdb/pull/63 where an issue is abandoned :trollface:

PR #66, opened 2017-11-15
Add configurable flush retry logic to AbstractBatch

luiscleto:
Currently, whenever the AbstractBatch.flush() method encounters an error, it fails immediately. This was because we do not know what caused the error and it might not make sense to retry.
However, some projects have run into errors while flushing that can be worked around simply by retrying the flush, avoiding having to write the failed batch to a dump file.
This pull request adds support for such workarounds by creating a configurable retry mechanism (which is disabled by default for retro-compatibility).

### DIFF
--- a/src/main/java/com/feedzai/commons/sql/abstraction/batch/AbstractBatch.java
+++ b/src/main/java/com/feedzai/commons/sql/abstraction/batch/AbstractBatch.java
@@ -55,6 +55,14 @@ public abstract class AbstractBatch implements Runnable {
      * Constant {@link FailureListener} representing the NO OP operation.
      */
     public final static FailureListener NO_OP = rowsFailed -> {};
+    /**
+     * Constant representing that no retries should be attempted on batch flush failures.
+     */
+    public final static int NO_RETRY = 0;
+    /**
+     * Constant representing the default time interval (milliseconds) to wait between batch flush retries.
+     */
+    public final static long DEFAULT_RETRY_INTERVAL = 300L;
 
     /**
      * The dev Marker.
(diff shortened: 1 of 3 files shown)


src/main/java/com/feedzai/commons/sql/abstraction/batch/AbstractBatch.java
```diff
@@ -139,16 +153,22 @@
      * @param batchTimeout         The batch timeout.
      * @param maxAwaitTimeShutdown The maximum await time for the batch to shutdown.
      * @param failureListener      The listener that will be invoked whenever some batch operation fail to persist.
+     * @param maxFlushRetries      The number of times to retry a batch flush upon failure. Defaults to
+     *                             {@value NO_RETRY}. When set to 0, no retries will be attempted.
+     * @param flushRetryDelay      The time interval (milliseconds) to wait between batch flush retries. Defaults to
```
kikevcosta: Doesn't make much sense to mention a default in a constructor that's forcing the user to pass the value. You can add this information in the other constructor.

src/main/java/com/feedzai/commons/sql/abstraction/batch/AbstractBatch.java
```diff
@@ -281,40 +325,71 @@ public void flush() {
...
-             *
-             * The idea here is to hand the entries that could not be added and continue right where we were.
-             *
-             * So, temp will only have the entries until it failed, this can be all the entries or until de.addBatch() failed;
-             * and buffer will have the events right after the failure.
-             */
-            onFlushFailure(temp.toArray(new BatchEntry[temp.size()]));
+        } catch (final Exception e) {
+            if (this.maxFlushRetries > 0) {
+                logger.debug(dev, "[{}] Error occurred while flushing. Retrying.", name, e);
```
kikevcosta: This should be a warn. Even though we're going to retry and everything might possibly be recovered, it is important for sysadmins to know that a failure happened.

src/main/java/com/feedzai/commons/sql/abstraction/batch/DefaultBatch.java
```diff
@@ -55,10 +56,30 @@ protected DefaultBatch(final DatabaseEngine de, final String name, final int bat
...
+     * @param name                 The batch name.
+     * @param batchSize            The batch size.
+     * @param batchTimeout         The timeout.
+     * @param maxAwaitTimeShutdown The maximum await time for the batch to shutdown.
+     * @param listener             The listener that will be invoked when batch fails to persist at least one data row.
+     * @param maxFlushRetries      The number of times to retry a batch flush upon failure. Defaults to
+     *                             {@value NO_RETRY}. When set to 0, no retries will be attempted.
+     * @param flushRetryDelay      The time interval (milliseconds) to wait between batch flush retries. Defaults to
+     *                             {@value DEFAULT_RETRY_INTERVAL}.     */
+    protected DefaultBatch(final DatabaseEngine de, final String name, final int batchSize, final long batchTimeout,
```
kikevcosta: Missing a `@since` and the javadoc end should be on its own line.

src/main/java/com/feedzai/commons/sql/abstraction/batch/AbstractBatch.java
```diff
@@ -281,40 +324,71 @@ public void flush() {
...
+                    // If the connection was established, we might need a rollback.
+                    if (de.checkConnection() && de.isTransactionActive()) {
+                        de.rollback();
+                    }
+
+                    // begin the transaction before the addBatch calls in order to force the retry
+                    // of the connection if the same was lost during or since the last batch. Otherwise
+                    // the addBatch call that uses a prepared statement will fail
+                    de.beginTransaction();
+
```
dspereira004: We are repeating the same code the first time we are trying to execute the batches and on the repeat. We should extract that (also the method is already quite large).

PR #68, opened 2018-02-02
Prevent batch executions from being suspended due to callback failures

bugflux:
This change prevents an issue where, if a call to `onFlushFailure` throws an exception, subsequent executions of the timer task will be suspended.

Additionally, it adds additional protection to the run method, to prevent regressions in changes to the `flush` method.

### DIFF
--- a/src/main/java/com/feedzai/commons/sql/abstraction/batch/AbstractBatch.java
+++ b/src/main/java/com/feedzai/commons/sql/abstraction/batch/AbstractBatch.java
@@ -185,7 +185,17 @@ public abstract class AbstractBatch implements Runnable {
      * Starts the timer task.
      */
     protected void start() {
-        scheduler.scheduleAtFixedRate(this, 0, batchTimeout + salt, TimeUnit.MILLISECONDS);
+        // if a periodic execution throws an exception, future executions are suspended,
+        // this task wraps the call in a try-catch block to prevent that. Errors are still propagated.
+        final Runnable resilientTask = () -> {
+            try {
+                run();
+            } catch (final Exception e) {
+                logger.error("[{}] Error during timeout-initiated flush", name, e);
+            }
+        };
+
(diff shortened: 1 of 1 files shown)


rfer:
@bugflux I'm having hard time justifying these changes. This is a problem upstream, not here.

It makes sense to propagate these exceptions. Having it like this is very specific to your use-case. What if the upstream logic is expecting some exception to be thrown in case of failure?

In truth, if the run() method can throw an exception, it should probably not even be runnable at all. 
I agree with the logging, as it allow us to have more fine grain insight of what went wrong, given the lack of a better alternative as there are no specific exceptions nor error codes. But the exceptions should be propagated, and have your upstream code wrap the call to run().

bugflux:
Hi Ricardo,

I’m ok with adjusting however you like.

However, I think the side effect of the batch being suspended as a result
of a failed call to onFlushFailure either needs to be prevented or clearly
supported/documented.

bugflux:
As for the wrap to “run”, I tend to agree.

I considered making the timer task a lambda that calls run while catching exceptions.

rfer:
@bugflux I misinterpreted this. I forgot about the Scheduled Executor on the AbstractBatch itself.

What I said still stands. And in light of that, we should handle this on the submit to AbstractBatch's Scheduled Executor.

bugflux:
I am torn, let me explain what my thought process was and I can adjust the pull request if you want me to:

I agree with (1) not suppressing the exception in the onFlushFailure call.
I also agree with (2) handling this at the submission to the scheduled execution

However, I think that the `run()` method, despite being public, doesn't seem part of the "API" of AbstractBatch, as its logging clearly states that it's running in the scope of a scheduled trigger.

So should we (3.1) make the AbstractBatch no longer a Runnable, or (3.2) embrace the `run()` method as THE place where the logic for the scheduled execution takes place?

PR #71, opened 2018-03-08
ref PULSEDEV-22452 pdb: Have DBEngine close all created prepared stat…

dfialho:
This pull request tries to solve the following issues:

- The base implementation of the DatabaseEngine keeps three prepared statements per entry for inserting. However, one of them (insertWithAutoInc) is not closed when the database is closed. 
- Furthermore, although the DatabaseEngine provides a method to create prepared statements, it does not closed them before closing the connection.


src/main/java/com/feedzai/commons/sql/abstraction/engine/AbstractDatabaseEngine.java
```diff
@@ -398,20 +398,12 @@ public synchronized void close() {
...
-                                try {
-                                    insertReturning.close();
-                                } catch (Exception e) {
-                                    logger.trace("Could not close prepared statement.", e);
-                                }
+                            try {
+                                me.getValue().close();
+                            } catch (Exception e) {
+                                // Should never be thrown
+                                logger.trace("Could not close insert statements.", e);
```
nmldiegues: This is an erroneous situation, but from which we are going to gracefully exit, so let's use WARN logging instead. Also, maybe refer that the insert statements are for a "MappedEntity" (i.e, they are not "custom" prepared statements as the one below)

src/main/java/com/feedzai/commons/sql/abstraction/engine/AbstractDatabaseEngine.java
```diff
@@ -422,6 +413,11 @@ public synchronized void close() {
                 }
             }
 
+            for (PreparedStatementCapsule ps : stmts.values()) {
+                ps.ps.close();
```
henriquevcosta: Consider wrapping in a try/catch so that we try to close as many as possible (and avoid misleading the user with a log message stating "Unable to close connection"

henriquevcosta: Also, since you're here, could you please add the exception instance to the log message in line 424?

src/main/java/com/feedzai/commons/sql/abstraction/engine/AbstractDatabaseEngine.java
```diff
@@ -398,20 +398,11 @@ public synchronized void close() {
...
-                            }
 
-                            if (insertReturning != null) {
-                                try {
-                                    insertReturning.close();
-                                } catch (Exception e) {
-                                    logger.trace("Could not close prepared statement.", e);
-                                }
+                            try {
+                                me.getValue().close();
```
henriquevcosta: This is only closing the insert statements when the  `properties.isSchemaPolicyCreateDrop()` property is set, which we agreed was strange. 
From our understanding the entity and its statements should be closed regardless, so this if should do an entity close and drop, and an else should do the close. Makes sense?

src/main/java/com/feedzai/commons/sql/abstraction/engine/AbstractDatabaseEngine.java
dfialho: Addressed in b449999

src/main/java/com/feedzai/commons/sql/abstraction/engine/AbstractDatabaseEngine.java
dfialho: Addressed in b9c7fc0

src/main/java/com/feedzai/commons/sql/abstraction/engine/AbstractDatabaseEngine.java
dfialho: Addressed in 7ade239

dfialho:
About the tests, I've run tests with PostgreSQL. All 186 tests passed. However, the logs show some exceptions. Is this "normal"?

I intent to test with the other vendors as well. The required docker images are taking a long time to download. I'll report back after I'm able to run the tests with all vendors.

dspereira004:
Currently, you can test each vendor individually with the option -P<vendor> which is what @dfialho is doing.

About the tests, I can't tell you for sure about all of them, but some of those test that we catch exceptions. In those cases, it is normal to appear exceptions in the test logs.

src/main/java/com/feedzai/commons/sql/abstraction/engine/AbstractDatabaseEngine.java
```diff
@@ -398,34 +398,48 @@ public synchronized void close() {
...
                     } catch (DatabaseEngineException ex) {
                         logger.debug(String.format("Failed to drop entity '%s'", me.getValue().getEntity().getName()), ex);
                     }
                 }
+
+            } else {
+                entities.forEach((key, mappedEntity) -> closeMappedEntity(mappedEntity));
+            }
+
+            for (PreparedStatementCapsule preparedStatement : stmts.values()) {
```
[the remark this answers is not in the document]
dfialho: Can you clarify when is it appropriate to include "final"?

- You're suggesting to include "final" in the for loop, which I totally understand. However, the loop above, which was already there, does not include the "final". Should I add "final" there as well?
- What about the catch clauses? Throughout the code there are catch clauses which include the "final" and other which don't. What should I do in those cases?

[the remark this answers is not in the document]
dspereira004: Appropriate is basically in every variable declaration that you know will never change (for primitive types, their value will never change; for objects, their reference will never change).

As a rule of thumb, you should put them in all situations that I stated. There are a lot of those missing throughout the code because we are not enforcing this rule (but mostly because it's older code). But for now, just add those to your code only. Also, Intellij inspection should highlight where "final"s are missing.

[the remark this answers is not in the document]
dfialho: Makes sense to me. Thank you for clarifying.

dfialho:
I'll wait for @henriquevcosta's thoughts on whether or not to flush regardless of the schema policy before committing new changes.

henriquevcosta:
I agree that is what makes sense, yes.

dfialho:
The mapped entity holds 3 statements. However, only two of them are flushed. Shouldn't flushing also extended to the 'insertWithAutoInc' statement?

henriquevcosta:
This is looking good. Please do flush the `insertWithAutoInc` statement, yes. How are we on testing this change?

dfialho:
I've run the currently existing tests and they all passed. However, none of the tests ensures the statements are closed correctly. I'll work on creating some tests for this.

dfialho:
@dspereira004 and @henriquevcosta, thoughts on the recent changes please :smile:

src/main/java/com/feedzai/commons/sql/abstraction/engine/AbstractDatabaseEngine.java
```diff
@@ -633,13 +643,32 @@ public synchronized void dropEntity(String entity) throws DatabaseEngineExceptio
...
     }
 
+    /**
+     * Drops all entities associated with this engine.
+     *
+     * @since 2.1.13
+     */
+    private void dropAllEntities() {
+
+        for (MappedEntity mappedEntity : ImmutableList.copyOf(entities.values())) {
```
henriquevcosta: `mappedEntity` can be final, and you have a few blank lines in this method that don't really improve legibility.

src/test/java/com/feedzai/commons/sql/abstraction/engine/impl/abs/EngineCloseTest.java
```diff
@@ -0,0 +1,148 @@
...
+        this.schemaPolicy = schemaPolicy;
+    }
+
+    @BeforeClass
+    public static void initStatic() {
+        ((Logger) LoggerFactory.getLogger(Logger.ROOT_LOGGER_NAME)).setLevel(Level.TRACE);
+    }
+
+    @Before
+    public void init() throws DatabaseFactoryException {
```
henriquevcosta: `setUp` is the standard name :)

src/main/java/com/feedzai/commons/sql/abstraction/engine/AbstractDatabaseEngine.java
```diff
@@ -57,14 +58,7 @@
...
 import java.sql.Types;
-import java.util.ArrayList;
-import java.util.HashMap;
-import java.util.HashSet;
-import java.util.LinkedHashMap;
-import java.util.List;
-import java.util.Map;
-import java.util.Properties;
-import java.util.Set;
+import java.util.*;
```
dspereira004: Could you please undo this import, meaning: import only the required classes? Also, do that on other classes where this also happened.
I think you have your Intellij IDEA configured to automatically import using "*" whenever you have "n" number of imports from the same package. "n" seems to be 8 on your case: I think you should increase this value a lot so this doesn't happen.

src/test/java/com/feedzai/commons/sql/abstraction/engine/EngineTestUtils.java
```diff
@@ -0,0 +1,50 @@
+/*
+ * Copyright 2014 Feedzai
```
dspereira004: 2018 and on the other one too

src/test/java/com/feedzai/commons/sql/abstraction/engine/impl/abs/EngineCloseTest.java
```diff
@@ -0,0 +1,148 @@
...
+        final Collection<DatabaseConfiguration> configurations = DatabaseTestUtil.loadConfigurations();
+        final Collection<String> schemaPolicies = Arrays.asList(
+                "drop-create",
+                "create-drop",
+                "create",
+                "none"
+        );
+
+        final ArrayList<Object[]> data = new ArrayList<>();
+        for (DatabaseConfiguration configuration : configurations) {
```
dspereira004: Missing finals on these two lines.

src/test/java/com/feedzai/commons/sql/abstraction/engine/impl/abs/EngineCloseTest.java
```diff
@@ -0,0 +1,148 @@
...
+    // Parameters
+    protected DatabaseConfiguration config;
+    protected String schemaPolicy;
+
+    @Parameterized.Parameters
+    public static Collection<Object[]> data() throws Exception {
+
+        final Collection<DatabaseConfiguration> configurations = DatabaseTestUtil.loadConfigurations();
+        final Collection<String> schemaPolicies = Arrays.asList(
+                "drop-create",
```
dspereira004: This is OK for now and especially for this pull-request, but in the future, we should create an Enum for this instead of using strings everywhere (and then using it across the project).

src/test/java/com/feedzai/commons/sql/abstraction/engine/impl/abs/EngineCloseTest.java
```diff
@@ -0,0 +1,148 @@
...
+
+    @BeforeClass
+    public static void initStatic() {
+        ((Logger) LoggerFactory.getLogger(Logger.ROOT_LOGGER_NAME)).setLevel(Level.TRACE);
+    }
+
+    @Before
+    public void setUp() throws DatabaseFactoryException {
+        properties = new Properties() {
+
```
dspereira004: extra line

src/test/java/com/feedzai/commons/sql/abstraction/engine/EngineTestUtils.java
```diff
@@ -0,0 +1,50 @@
...
+import static com.feedzai.commons.sql.abstraction.ddl.DbColumnType.*;
+import static com.feedzai.commons.sql.abstraction.dml.dialect.SqlBuilder.dbEntity;
+
+/**
+ * Utils for using in {@link DatabaseEngine} tests.
+ *
+ * @author David Fialho (david.fialho@feedzai.com)
+ * @since 2.1.13
+ */
+public interface EngineTestUtils {
```
dspereira004: Any reason why you chose an interface instead of a class?

src/test/java/com/feedzai/commons/sql/abstraction/engine/impl/abs/EngineGeneralTest.java
```diff
@@ -3495,4 +3495,35 @@ void close() {
...
         );
     }
+
+    /**
+     * Test that closing a database engine a 'create-drop' policywith multiple entities closes all insert statements associated with each
+     * entity, regardless of the schema policy used.
+     *
+     * Each entity is associated with 3 prepared statements. This test ensures that 3 PSs per entity are closed.
+     *
+     * @throws DatabaseEngineException    If something goes wrong while adding an entity to the engine.
```
dspereira004: I think there is a "@throws" missing here.

src/test/java/com/feedzai/commons/sql/abstraction/engine/impl/abs/EngineCloseTest.java
dfialho: Totally agree :+1:

src/test/java/com/feedzai/commons/sql/abstraction/engine/EngineTestUtils.java
dfialho: I used an interface because this is supposed to include static methods. As far as I know, using a class for utilities just adds unnecessary boilerplate code. I might be wrong, though. Is there any reason to use a class instead of an interface for this purpose?

dspereira004: That boilerplate code would just be a private constructor to prevent instantiation: nothing else. Let me put it this way: do you imagine that interface be implemented by several classes? Or are you thinking that in the future we are going to leverage polymorphism and use implementations of EngineTestUtil?

If you answer yes to either question, that perhaps my judgment is wrong. 

Otherwise, from the name of the interface and its comments, you are creating something to helps you do other stuff which is not directly related to "objects" and "polymorphism". In that case, you don't want instantiation (and therefore you don't want to create objects). Right now I can to that by implementing that interface. For a "true" utility "class", you could just create a "public final class EngineTestUtils" and create a private constructor (which will prevent creating instances or extend that class). 

Therefore, the changes would be:
- change "interface" to "final class"
- create private constructor
- make that static method public.

Also, I think that you still have imports with "*". Could you please take a look at that?

dfialho: You make a good point about a subclass/implementation still being able to be instantiated.
I'll make the proposed changes. 

Thank you for taking the time to answer this thoroughly.

dspereira004:
Thanks for this @dfialho

dfialho:
Thank you for your input @henriquevcosta and @dspereira004

PR #72, opened 2018-03-09
Add "pdb.compress_lobs" option to allow the compression of BLOBS and CLOBS

dspereira004:
This commit adds a new option to compress LOB columns. Currently, only the
Oracle implementation is supported.

### DIFF
--- a/src/main/java/com/feedzai/commons/sql/abstraction/engine/impl/OracleEngine.java
+++ b/src/main/java/com/feedzai/commons/sql/abstraction/engine/impl/OracleEngine.java
@@ -52,6 +52,7 @@ import java.util.Optional;
 import static com.feedzai.commons.sql.abstraction.util.StringUtils.md5;
 import static com.feedzai.commons.sql.abstraction.util.StringUtils.quotize;
 import static java.lang.String.format;
+import static java.util.stream.Collectors.joining;
 import static org.apache.commons.lang3.StringUtils.join;
 
 /**
@@ -86,6 +87,11 @@ public class OracleEngine extends AbstractDatabaseEngine {
      * Foreign key already exists
      */
     public static final String FOREIGN_ALREADY_EXISTS = "ORA-02275";
+
+    /**
+     * Secure file LOBS cannot be used in non-ASSM (Automatic Segment Space Management) tablespace.
+     */
+    private static final String SECUREFILE_NOT_ASSM = "ORA-43853";
     /**
      *  Double instance for 0.0, so no Double instance is created whenever 0.0 is necessary.
      *
@@ -233,7 +239,7 @@ public class OracleEngine extends AbstractDatabaseEngine {
         if (properties.getSchema() != null && properties.getSchema().length() > 0 &&
                         !properties.getUsername().equals(properties.getSchema())) {
             // Set connection schema if one is defined and not the same as the user name
-            Statement stmt = conn.createStatement();
+            final Statement stmt = conn.createStatement();
             stmt.execute("ALTER SESSION SET CURRENT_SCHEMA=" + properties.getSchema());
             stmt.close();
         }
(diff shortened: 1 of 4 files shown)


src/main/java/com/feedzai/commons/sql/abstraction/engine/impl/OracleEngine.java
```diff
@@ -268,13 +274,51 @@ protected void setTransactionIsolation() throws SQLException {
...
+        final String defaultCreateTableStatement = getCreateTableStatement(entity);
+
+        // Check for extra properties
+        final String createTableCompressLOBS = getCompressLobsStatements(entity);
+
+        final String createTableStatement = defaultCreateTableStatement + createTableCompressLOBS;
+
+        for (final String createTable : ImmutableList.of(createTableStatement, defaultCreateTableStatement)) {
+            try (
+                    Statement statement = conn.createStatement()
```
jmf-tls: How about putting this in a single line, with a _final_
`            try (final Statement statement = conn.createStatement()) {`
(the statement is not being closed...)

src/main/java/com/feedzai/commons/sql/abstraction/engine/impl/OracleEngine.java
```diff
@@ -268,13 +274,51 @@ protected void setTransactionIsolation() throws SQLException {
...
+            ) {
+                statement.executeUpdate(createTable);
+                break; // If we got here, then we successfully created the table
+            } catch (final SQLException ex) {
+                if (ex.getMessage().startsWith(NAME_ALREADY_EXISTS)) {
+                    logger.debug(dev, "'{}' is already defined", entity.getName());
+                    handleOperation(new OperationFault(entity.getName(), OperationFault.Type.TABLE_ALREADY_EXISTS), ex);
+                    break; // Name already exists, we cannot do anything so we exit
+                } else if (ex.getMessage().startsWith(SECUREFILE_NOT_ASSM)) {
```
jmf-tls: another _if_ should be added here for the case where SecureFiles are allowed, but LOB compression is not.
This is the corresponding error code:
> ORA-00439: feature not enabled: Advanced Compression

and if this happens, table should be created without compressed LOBs, just like the case for SECUREFILE_NOT_ASSM

src/main/java/com/feedzai/commons/sql/abstraction/engine/impl/OracleEngine.java
```diff
@@ -233,7 +239,7 @@ public void connect() throws Exception {
         if (properties.getSchema() != null && properties.getSchema().length() > 0 &&
                         !properties.getUsername().equals(properties.getSchema())) {
             // Set connection schema if one is defined and not the same as the user name
-            Statement stmt = conn.createStatement();
+            final Statement stmt = conn.createStatement();
```
jmf-tls: how about a try-with-resources, and drop the _stmt.close();_ below?

src/main/java/com/feedzai/commons/sql/abstraction/engine/impl/OracleEngine.java
```diff
@@ -268,13 +274,51 @@ protected void setTransactionIsolation() throws SQLException {
...
+            try (
+                    Statement statement = conn.createStatement()
+            ) {
+                statement.executeUpdate(createTable);
+                break; // If we got here, then we successfully created the table
+            } catch (final SQLException ex) {
+                if (ex.getMessage().startsWith(NAME_ALREADY_EXISTS)) {
+                    logger.debug(dev, "'{}' is already defined", entity.getName());
+                    handleOperation(new OperationFault(entity.getName(), OperationFault.Type.TABLE_ALREADY_EXISTS), ex);
+                    break; // Name already exists, we cannot do anything so we exit
```
jmf-tls: nothing to do after this... could be a _return;_ instead?

src/test/java/com/feedzai/commons/sql/abstraction/engine/impl/oracle/OracleEngineSchemaTest.java
```diff
@@ -97,10 +112,101 @@ public void testSystemGeneratedColumns() throws Exception {
...
+     *
+     * @throws Exception if anything goes wrong with the test
+     */
+    @Test
+    public void testCompressLobs() throws Exception {
+        final String tablespace = "TEST_TABLESPACE";
+        final String tableName = "TEST_TABLE";
+        final String tableUserLobs = "USER_LOBS";
+
+        final String tableId = "ID";
```
jmf-tls: maybe these 3 should be called _column*_ or _*column_
instead of _table*_

src/test/java/com/feedzai/commons/sql/abstraction/engine/impl/oracle/OracleEngineSchemaTest.java
```diff
@@ -97,10 +112,101 @@ public void testSystemGeneratedColumns() throws Exception {
...
+        final String tableClob = "CLOB_COLUMN";
+
+        final String compression = "COMPRESSION";
+        final String secureFile = "SECUREFILE";
+
+        properties.setProperty(COMPRESS_LOBS, Boolean.toString(true));
+
+        final DatabaseEngine engine = DatabaseFactory.getConnection(properties);
+
+        createUserTablespace(tablespace);
```
jmf-tls: not sure if...
could these _create/updateUserTablespace_ use "engine" intead of creating another one in each method
and maybe both could be joined into just 1 method?

src/main/java/com/feedzai/commons/sql/abstraction/engine/impl/OracleEngine.java
```diff
@@ -292,31 +336,29 @@ protected void createTable(final DbEntity entity) throws DatabaseEngineException
...
+     * @return the extra properties.
+     * @since 2.1.13
+     */
+    private String getCompressLobsStatements(final DbEntity entity) {
+        if (!this.properties.shouldCompressLobs()) {
+            return "";
+        }
 
-        logger.trace(createTableStatement);
+        final String defaultLob = " LOB (\"%s\") STORE AS SECUREFILE (COMPRESS HIGH CACHE LOGGING)";
```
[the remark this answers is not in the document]
dspereira004: I definitely agree with you: I will change for MEDIUM

src/test/java/com/feedzai/commons/sql/abstraction/engine/impl/oracle/OracleEngineSchemaTest.java
dspereira004: Yes, they can. I will do that.

src/test/java/com/feedzai/commons/sql/abstraction/engine/impl/oracle/OracleEngineSchemaTest.java
dspereira004: Yup, makes sense

src/main/java/com/feedzai/commons/sql/abstraction/engine/impl/OracleEngine.java
dspereira004: It was meant to be a return: somehow I put a break there (facepalm).

src/main/java/com/feedzai/commons/sql/abstraction/engine/impl/OracleEngine.java
dspereira004: Right, let's clean up code whenever we have the opportunity. Thanks for the remark

src/main/java/com/feedzai/commons/sql/abstraction/engine/impl/OracleEngine.java
dspereira004: Really nice catch! Thanks for that, I will update this soon!

src/main/java/com/feedzai/commons/sql/abstraction/engine/impl/OracleEngine.java
dspereira004: I'm currently using try-with-resources so the statement is being closed automatically. Anyway, I will do a few changes in this pull-request :)

jmf-tls: uups
Yes, the "statement is not being closed currently" was meant for line 242 below.
My "copy-paste comment" didn't work very well

Issue #75, opened 2018-04-18
SQL Server order by with table name - The multi-part identifier could not be bound


jrrbarros:
When using an SQL Query which contains an ORDER BY clause to a TABLE.COLUMN column, I get this error: The multi-part identifier could not be bound.
This only happens with SQL Server which does not support this syntax.

jrrbarros:
Example of a problematic query:
`final Query query = select(all())
                .from(table("USER"))
                .orderby(ImmutableList.of(column("USER", "AGE").asc()))
                .limit(1)
                .offset(1);`

PR #76, opened 2018-04-19
ref #75: SQL Server order by with table name - The multi-part identifier could not be bound

jrrbarros:
#75 
SQL Server was throwing errors when using multipart identifiers inside ORDER BY clauses. This only happen when we're using paginated requests, with limit and offset values, which requires sub-queries inside the main query.
This patch verifies if table name is present in this case. If this condition is true, we remove it and only leave a column with its name.


jrrbarros:
Refactored table and columns names. ASC and DESC are better as is (for now).

PR #80, opened 2018-05-11
fixes #73 - Fix schemas

### LINKED ISSUE #73: "schema" property isn't respected
Several problems were found regarding the usage of "schema" PDB property:
- it is almost completely ignored (it seems only Oracle engine really sets current schema, other engines only used schema to get metadata and/or translate functions)
- the methods to get metadata (_getMetadata_, _getEntities_) return results that don't match the current schema -  for example, if schema is _not_ set in PDB properties, _updateEntity_ will do the following
    - check if the entity exists (since "schema" was not set in properties, it will try to find the entity in all the schemas)
    - if it doesn't exist, add it; at this point, if the entity exists on another schema, it won't be added to the current schema - inserting a new entry will either fail (if the engine doesn't look in other schemas if the entity is missing in the current one), or it will use the other schema (which translates into writing to a table that's not the one in the current default schema)
    - considering a migration scenario, where some data from a table is to be copied to other table with the same name, in the same database, but different schema: with a schema policy drop-create, _updateEntity_ might even try to _**drop the entity on the wrong schema**_

jmf-tls:
This addresses PULSEDEV-23140, where Oracle engine had some problems when schema property was defined and was different from the user, and all other engines weren't respecting the schema property at all.

The tests were updated accordingly.

### DIFF
--- a/src/main/java/com/feedzai/commons/sql/abstraction/engine/impl/DB2Engine.java
+++ b/src/main/java/com/feedzai/commons/sql/abstraction/engine/impl/DB2Engine.java
@@ -148,7 +148,7 @@ public class DB2Engine extends AbstractDatabaseEngine {
                     default:
                         ps.setObject(i, val);
                 }
-            } catch (Exception ex) {
+            } catch (final Exception ex) {
                 throw new DatabaseEngineException("Error while mapping variables to database", ex);
             }
 
@@ -192,7 +192,7 @@ public class DB2Engine extends AbstractDatabaseEngine {
         try {
             s = conn.createStatement();
             s.executeUpdate(createTableStatement);
-        } catch (SQLException ex) {
+        } catch (final SQLException ex) {
             if (ex.getMessage().startsWith(NAME_ALREADY_EXISTS)) {
                 logger.debug(dev, "'{}' is already defined", entity.getName());
                 handleOperation(new OperationFault(entity.getName(), OperationFault.Type.TABLE_ALREADY_EXISTS), ex);
@@ -204,7 +204,7 @@ public class DB2Engine extends AbstractDatabaseEngine {
                 if (s != null) {
                     s.close();
                 }
-            } catch (Exception e) {
+            } catch (final Exception e) {
                 logger.trace("Error closing statement.", e);
             }
         }
@@ -258,7 +258,7 @@ public class DB2Engine extends AbstractDatabaseEngine {
             logger.trace(reorg);
             s = conn.createStatement();
             s.executeUpdate(reorg);
-        } catch (SQLException ex) {
+        } catch (final SQLException ex) {
             if (ex.getMessage().startsWith(TABLE_CAN_ONLY_HAVE_ONE_PRIMARY_KEY)) {
                 logger.debug(dev, "'{}' already has a primary key", entity.getName());
                 handleOperation(new OperationFault(entity.getName(), OperationFault.Type.PRIMARY_KEY_ALREADY_EXISTS), ex);
@@ -270,7 +270,7 @@ public class DB2Engine extends AbstractDatabaseEngine {
                 if (s != null) {
                     s.close();
                 }
-            } catch (Exception e) {
+            } catch (final Exception e) {
                 logger.trace("Error closing statement.", e);
             }
         }
@@ -346,7 +346,7 @@ public class DB2Engine extends AbstractDatabaseEngine {
             try {
                 s = conn.createStatement();
                 s.executeUpdate(statement);
-            } catch (SQLException ex) {
+            } catch (final SQLException ex) {
                 if (ex.getMessage().startsWith(NAME_ALREADY_EXISTS)) {
                     logger.debug(dev, "'{}' is already defined", idxName);
                     handleOperation(new OperationFault(entity.getName(), OperationFault.Type.INDEX_ALREADY_EXISTS), ex);
@@ -358,7 +358,7 @@ public class DB2Engine extends AbstractDatabaseEngine {
                     if (s != null) {
                         s.close();
                     }
-                } catch (Exception e) {
+                } catch (final Exception e) {
                     logger.trace("Error closing statement.", e);
                 }
             }
@@ -402,7 +402,7 @@ public class DB2Engine extends AbstractDatabaseEngine {
             try {
                 s = conn.createStatement();
                 s.executeUpdate(statement);
-            } catch (SQLException ex) {
+            } catch (final SQLException ex) {
                 if (ex.getMessage().startsWith(NAME_ALREADY_EXISTS)) {
                     logger.debug(dev, "'{}' is already defined", sequenceName);
                     handleOperation(new OperationFault(entity.getName(), OperationFault.Type.SEQUENCE_ALREADY_EXISTS), ex);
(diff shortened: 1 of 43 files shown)


src/main/java/com/feedzai/commons/sql/abstraction/engine/impl/H2Translator.java
```diff
@@ -86,11 +86,7 @@ public String translate(Function f) {
         }
 
         // if it is a user-defined function
-        if (f.isUDF() && properties.isSchemaSet()) {
-            return properties.getSchema() + "." + function + "(" + expTranslated + ")";
-        } else {
-            return function + "(" + expTranslated + ")";
-        }
+        return function + "(" + expTranslated + ")";
```
dspereira004: This means that the getSchema was never needed? Also, if you're gonna let this as-is, then the comment doesn't make sense anymore. The same thing applies to the other DB implementations.

src/main/java/com/feedzai/commons/sql/abstraction/engine/impl/MySqlEngine.java
```diff
@@ -764,6 +768,63 @@ protected boolean checkConnection(final Connection conn) {
         }
     }
 
+    @Override
+    protected String getSchema() throws DatabaseEngineException {
+        try {
+            return this.conn.getCatalog();
```
dspereira004: Why **getCatalog** instead of **getSchema** (the default method call on the superclasse)?

src/main/java/com/feedzai/commons/sql/abstraction/engine/impl/OracleEngine.java
```diff
@@ -954,31 +944,21 @@ protected boolean checkConnection(final Connection conn) {
...
     }
 
-    /**
-     * Gets the schema. This implementation uses the username when the schema is null.
-     *
-     * @return The schema or the username when the first is null.
-     */
-    @Override
-    protected String getSchema() {
-        return Optional.ofNullable(super.getSchema()).orElse(properties.getUsername());
```
dspereira004: What happens now if we call getSchema and it is null?

src/test/java/com/feedzai/commons/sql/abstraction/engine/impl/abs/AbstractEngineSchemaTest.java
```diff
@@ -639,9 +741,65 @@ protected EntityEntry createSpecialValueEntry(final Object columnValue) {
...
+     * @throws DatabaseEngineException  If anything goes wrong creating the schema.
+     * @throws DatabaseFactoryException If anything goes wrong connecting to the database.
+     * @since 2.1.13
+     */
+    public void dropCreateTestSchema() throws DatabaseEngineException, DatabaseFactoryException {
+        try (final DatabaseEngine engine = DatabaseFactory.getConnection(properties)) {
+            try {
+                dropSchema(engine, getTestSchema());
+            } catch (DatabaseEngineException e) {
+                // do nothing
```
dspereira004: Why not just remove the try-catch here and let the Exception being thrown upwards to the other methods?

src/main/java/com/feedzai/commons/sql/abstraction/engine/impl/DB2Engine.java
```diff
@@ -852,7 +852,45 @@ protected ResultIterator createResultIterator(Statement statement, String sql) t
...
+            // result needs to be trimmed because it has padding spaces
+            return schema == null ? null : schema.trim();
+        } catch (Exception e) {
+            throw new DatabaseEngineException("Could not get current schema", e);
+        }
+    }
+
+    @Override
+    protected void setSchema(final String schema) throws DatabaseEngineException {
+        final String findSchemaQuery = "SELECT count(*) FROM syscat.schemata WHERE SCHEMANAME = '" + schema + "'";
```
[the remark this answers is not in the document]
jmf-tls: Totally right, glad to have an extra pair of eyes looking at it!
Prepared statements don't even need values to be quotized, the driver will do the proper thing for us :+1:

src/main/java/com/feedzai/commons/sql/abstraction/engine/impl/DB2Engine.java
```diff
@@ -862,8 +900,11 @@ protected ResultIterator createResultIterator(Statement statement, String sql) t
 
             s = conn.createStatement();
             rsColumns = s.executeQuery(
-                String.format("SELECT NAME, COLTYPE,SCALE  FROM sysibm.SYSCOLUMNS WHERE tbname='%s' and TBCREATOR=UPPER('%s')",
-                    name, Optional.ofNullable(properties.getSchema()).orElse(properties.getUsername()))
+                String.format(
+                    "SELECT NAME, COLTYPE, SCALE FROM sysibm.SYSCOLUMNS WHERE tbname LIKE '%s' and TBCREATOR LIKE '%s'",
```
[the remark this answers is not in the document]
jmf-tls: The _getMetadata(final String name)_ method was calling
_java.sql.DatabaseMetaData.getColumns(String catalog, String schemaPattern, String tableNamePattern, String columnNamePattern)_
for most DB engines (DB2 was one of the exceptions).

Because of that, the "name" in reality was a table name pattern which might include wildcards.
Since DB2 driver doesn't implement that _getColumns_ method, this one should behave is if it did, by accepting patterns instead of only exact names for schema and table; for this to work, it is necessary to use LIKE.

src/main/java/com/feedzai/commons/sql/abstraction/engine/impl/MySqlEngine.java
jmf-tls: For MySql schema and catalog are the same thing.
However, for some reason _getSchema()_ doesn't work, just returns _null_ :(

src/main/java/com/feedzai/commons/sql/abstraction/engine/impl/OracleEngine.java
jmf-tls: With these changes, we have 3 sources for schema information:
* _properties.getSchema()_ from the PDB properties, which can be null (meaning "use the default schema")
* _AbstractDatabaseEngine.getSchema()_ which returns the current schema from the Connection -> the current schema name or null if there is none (there should always be a schema, which is why the _AbstractDatabaseEngine.connect()_ method will throw an exception if the connection schema is null)
* it is assumed that once set right after connecting to the DB, the current schema won't change; the _AbstractDatabaseEngine.currentSchema_ will contain the name of that schema, so that it can be used when needed without having to query the DB to get it

src/main/java/com/feedzai/commons/sql/abstraction/engine/impl/OracleEngine.java
```diff
@@ -241,27 +241,14 @@ protected void setParameterValues(final PreparedStatement ps, final int index, f
      * @since       2.1.4
      */
     private Object ensureNoUnderflow(Object val) {
-        if (val instanceof  Double) {
-            Double dblVal = (Double)val;
-            if (Math.abs(dblVal) <= 1.0e-131) {
-                val = ZERO;
+        if (val instanceof Double) {
```
dspereira004: Why not:

```java
if (val instanceof Double && Math.abs((Double) val) <= 1.0e-131) {
    return ZERO;
}

return val;
```

jmf-tls: :)

jmf-tls:
Since this is merged into master branch, I cleaned up history to have only 2 commits, both integrating all changes that were done following the review comments.

PR #87, opened 2018-06-05
fixes #86 - Fix Oracle and MySql engines to return correct results in getMetadata and getEntities

### LINKED ISSUE #86: Getting metadata/entities may not return the expected results in Oracle/MySql
**MySql**: just like _getMetadata_, the _getEntities_ method also has a "schema" parameter which is ignored since for this DB engine the "catalog" is considered the schema.

**Oracle**: the method _getMetadata(String name)_ was overriding the same method in _AbstractDatabaseEngine_, when the method _getMetadata(String schemaPattern, String tableNamePattern)_ should have been overridden instead (otherwise it would only work for the current schema).

The tests should be changed to verify that these methods are working as expected.

jmf-tls:
Summary:
MySql - just like getMetadata, the getEntities method also has a "schema"
parameter which is ignored since for this DB engine the "catalog" is
considered the schema - this commit puts the schema value in the
  "catalog" parameter.

Oracle - the method getMetadata(String name) was overriding the same
method in AbstractDatabaseEngine, when the method
  getMetadata(String schemaPattern, String tableNamePattern)
should have been overridden instead (otherwise it would only work for
the current schema) - this commit makes that change.

The tests were also changed to verify that these methods are working
as expected.

### DIFF
--- a/src/main/java/com/feedzai/commons/sql/abstraction/engine/impl/MySqlEngine.java
+++ b/src/main/java/com/feedzai/commons/sql/abstraction/engine/impl/MySqlEngine.java
@@ -19,6 +19,7 @@ import com.feedzai.commons.sql.abstraction.ddl.DbColumn;
 import com.feedzai.commons.sql.abstraction.ddl.DbColumnConstraint;
 import com.feedzai.commons.sql.abstraction.ddl.DbColumnType;
 import com.feedzai.commons.sql.abstraction.ddl.DbEntity;
+import com.feedzai.commons.sql.abstraction.ddl.DbEntityType;
 import com.feedzai.commons.sql.abstraction.ddl.DbFk;
 import com.feedzai.commons.sql.abstraction.ddl.DbIndex;
 import com.feedzai.commons.sql.abstraction.dml.dialect.Dialect;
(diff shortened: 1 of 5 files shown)


src/main/java/com/feedzai/commons/sql/abstraction/engine/impl/MySqlEngine.java
```diff
@@ -733,6 +734,53 @@ protected void dropReferringFks(DbEntity entity) throws DatabaseEngineException
...
+    @Override
+    public synchronized Map<String, DbEntityType> getEntities(final String schemaPattern) throws DatabaseEngineException {
+        final Map<String, DbEntityType> entities = new LinkedHashMap<>();
+        ResultSet rs = null;
+
+        try {
+            getConnection();
+
+            // get the entities
+            rs = conn.getMetaData().getTables(schemaPattern, null, "%", null);
```
dspereira004: missing "this" on the instance variable, please :)

dspereira004: Meh, just because of this difference (when compared to the same method of the abstract class), but yeah..

Could you please put the finals where necessary in this method?

jmf-tls:
I made the suggested changes and ran the tests for all 7 engine choices (all passed).

PR #94, opened 2018-11-30
Support Ceiling/Floor expressions.

### LINKED ISSUE #91: Support Ceiling/Floor expressions.
PostgresSQL Reference: https://www.postgresql.org/docs/11/functions-math.html

francisco-polaco:
Implementation of Ceiling/Floor expressions, which are Engine agnostic and useful when working with doubles.

This closes #91.

### DIFF
--- a/src/test/java/com/feedzai/commons/sql/abstraction/engine/impl/abs/EngineGeneralTest.java
+++ b/src/test/java/com/feedzai/commons/sql/abstraction/engine/impl/abs/EngineGeneralTest.java
@@ -76,6 +76,7 @@ import static com.feedzai.commons.sql.abstraction.dml.dialect.SqlBuilder.L;
 import static com.feedzai.commons.sql.abstraction.dml.dialect.SqlBuilder.all;
 import static com.feedzai.commons.sql.abstraction.dml.dialect.SqlBuilder.avg;
 import static com.feedzai.commons.sql.abstraction.dml.dialect.SqlBuilder.between;
+import static com.feedzai.commons.sql.abstraction.dml.dialect.SqlBuilder.ceiling;
 import static com.feedzai.commons.sql.abstraction.dml.dialect.SqlBuilder.coalesce;
 import static com.feedzai.commons.sql.abstraction.dml.dialect.SqlBuilder.column;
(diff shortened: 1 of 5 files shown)


src/test/java/com/feedzai/commons/sql/abstraction/engine/impl/abs/EngineGeneralTest.java
```diff
@@ -1282,6 +1284,50 @@ public void minTest() throws DatabaseEngineException {
...
+
+        for (int i = 0; i < 10; i++) {
+            entry.set("COL1", i);
+            engine.persist("TEST", entry
+                    .build());
+        }
+
+        List<Map<String, ResultColumn>> query = engine.query(select(floor(column("COL3")).alias("FLOOR")).from(table("TEST")));
+
+        assertEquals("result ok?", 2.0, query.get(0).get("FLOOR").toDouble(), 0.1);
```
marmelo: `0.1` is a very high delta.

Ideally there should be a static variable to be used in all tests but please check the STDDEV tests that are using `0.0001D` as delta.

src/test/java/com/feedzai/commons/sql/abstraction/engine/impl/abs/EngineGeneralTest.java
```diff
@@ -1282,6 +1284,50 @@ public void minTest() throws DatabaseEngineException {
...
+
+        for (int i = 0; i < 10; i++) {
+            entry.set("COL1", i);
+            engine.persist("TEST", entry
+                    .build());
+        }
+
+        List<Map<String, ResultColumn>> query = engine.query(select(ceiling(column("COL3")).alias("CEILING")).from(table("TEST")));
+
+        assertEquals("result ok?", 3.0, query.get(0).get("CEILING").toDouble(), 0.1);
```
marmelo: Same comment as above.

src/test/java/com/feedzai/commons/sql/abstraction/engine/impl/abs/EngineGeneralTest.java
marmelo: For reference, we usually do this:

```
public static final Double DELTA = 1e-7;
```

PR #95, opened 2018-11-30
Support CASE conditional expressions.

### LINKED ISSUE #26: Support CASE conditional expressions.
Please see PostgreSQL as a reference: http://www.postgresql.org/docs/current/static/functions-conditional.html

francisco-polaco:
Case when clauses.

This closes #26 .


src/main/java/com/feedzai/commons/sql/abstraction/dml/Case.java
```diff
@@ -0,0 +1,70 @@
+/*
+ * The copyright of this file belongs to Feedzai. The file cannot be
+ * reproduced in whole or in part, stored in a retrieval system,
+ * transmitted in any form, or by any means electronic, mechanical,
+ * photocopying, or otherwise, without the prior permission of the owner.
+ *
+ * (c) 2014 Feedzai, Strictly Confidential
```
marmelo: Typo, 2018.

src/main/java/com/feedzai/commons/sql/abstraction/dml/Case.java
```diff
@@ -0,0 +1,70 @@
...
+
+    /**
+     * Creates an empty case.
+     */
+    protected Case() {
+        this.whens = new ArrayList<>();
+    }
+
+    /**
+     * @return a new case.
```
marmelo: Missing Javadoc.

src/main/java/com/feedzai/commons/sql/abstraction/dml/Case.java
```diff
@@ -0,0 +1,70 @@
...
+     */
+    public static Case caseWhen() {
+        return new Case();
+    }
+
+    /**
+     * @param condition condition to verify.
```
marmelo: Missing Javadoc.

src/main/java/com/feedzai/commons/sql/abstraction/dml/Case.java
```diff
@@ -0,0 +1,70 @@
...
+        return caseWhen().when(condition, trueAction);
+    }
+
+    @Override
+    public String translate() {
+        return translator.translate(this);
+    }
+
+    /**
+     * @param condition condition to verify.
```
marmelo: Missing Javadoc.

src/main/java/com/feedzai/commons/sql/abstraction/dml/Case.java
```diff
@@ -0,0 +1,70 @@
...
+    public static Case caseWhen() {
+        return new Case();
+    }
+
+    /**
+     * @param condition condition to verify.
+     * @param trueAction action to be executed if the condition is true.
+     * @return a new "case when" that does the trueAction considering the condition.
+     */
+    public static Case caseWhen(final Expression condition, final Expression trueAction) {
```
marmelo: Is this method required given that we already have the `when(condition, action)` method?

marmelo: Nevermind, I like doing this when you just have 1 condition.

src/main/java/com/feedzai/commons/sql/abstraction/dml/Case.java
```diff
@@ -0,0 +1,70 @@
...
+    /**
+     * @param condition condition to verify.
+     * @param action action to be executed if the condition is true.
+     * @return this case.
+     */
+    public Case when(final Expression condition, final Expression action) {
+        whens.add(When.when(condition, action));
+        return this;
+    }
+
```
marmelo: Shouldn't you have a `whenElse(...)` method here?  
It seems strange having to instantiate a `CaseElse` if you want to have a else clause.

I would like doing something like:
```
caseWhen(condition, action)
    .when(condition, action)
    .when(condition, action)
    .whenElse(action);
```
Does it make sense?

I know we may not use `case` and `else` because they are keywords.  
But I am thinking that would be prettier to do something like:
```
when(condition, action)
    .when(condition, action)
    .when(condition, action)
    .otherwise(action);
```

src/main/java/com/feedzai/commons/sql/abstraction/dml/CaseElse.java
```diff
@@ -0,0 +1,66 @@
+/*
+ * The copyright of this file belongs to Feedzai. The file cannot be
+ * reproduced in whole or in part, stored in a retrieval system,
+ * transmitted in any form, or by any means electronic, mechanical,
+ * photocopying, or falseAction, without the prior permission of the owner.
+ *
+ * (c) 2014 Feedzai, Strictly Confidential
```
marmelo: Typo, 2018.

src/main/java/com/feedzai/commons/sql/abstraction/dml/CaseElse.java
```diff
@@ -0,0 +1,66 @@
...
+    public final Expression falseAction;
+
+    /**
+     * Creats a CaseElse.
+     *
+     * @param condition condition to verify.
+     * @param trueAction action to be executed if the condition is true.
+     * @param falseAction action to be executed if the condition is false.
+     */
+    private CaseElse(final Expression condition, final Expression trueAction, final Expression falseAction) {
```
marmelo: In line with the previous comment, maybe the `else` should only have an action.  
Seems that we are replicating the case logic here.

src/main/java/com/feedzai/commons/sql/abstraction/dml/dialect/SqlBuilder.java
```diff
@@ -519,6 +519,30 @@ public static View createView(final String name) {
...
+
+    /**
+     * Creates a case when else expression.
+     *
+     * @param condition The name of the view.
+     * @param trueAction The name of the view.
+     * @param falseAction The name of the view.
+     * @return The case when representation.
+     */
+    public static CaseElse caseElse(final Expression condition, final Expression trueAction,
```
marmelo: Again, I believe we should not have this, where else is only accessible through the above expression.

src/test/java/com/feedzai/commons/sql/abstraction/engine/impl/abs/EngineGeneralTest.java
```diff
@@ -1927,6 +1929,50 @@ public void betweenTest() throws DatabaseEngineException {
...
+        List<Map<String, ResultColumn>> query = engine.query(
+                select(all(), caseElse(eq(column("COL5"), k("teste")), k("LOL"), k("ROFL")).alias("case"))
+                        .from(table("TEST"))
+        );
+
+        assertEquals("COL5 must be LOL", "LOL", query.get(0).get("case").toString());
+        assertEquals("COL5 must be ROFL", "ROFL", query.get(1).get("case").toString());
+        assertEquals("COL5 must be ROFL", "ROFL", query.get(2).get("case").toString());
+        assertEquals("COL5 must be LOL", "LOL", query.get(3).get("case").toString());
+    }
```
marmelo: Missing test with more than 1 case.

src/main/java/com/feedzai/commons/sql/abstraction/dml/Case.java
francisco-polaco: Wouldn't this force us to write an if statement in the visitor, instead of using polymorphism?
I only created another class to avoid having an if, since the final part of the query changes a little.

Instead of `CASE WHEN ... THEN ... END `we are going to have `CASE WHEN ... THEN ... ELSE ... END`

Is it ok to check if otherwise is null?

src/main/java/com/feedzai/commons/sql/abstraction/dml/Case.java
```diff
@@ -0,0 +1,120 @@
...
+    /**
+     * Returns a new "case when" that returns true or false considering the condition.
+     *
+     * @param condition condition to verify.
+     * @return a new "case when" considering the condition.
+     */
+    public static Case caseWhen(final Expression condition) {
+        return caseWhen(condition, k(true), k(false));
```
marmelo: Why do we need this?

src/main/java/com/feedzai/commons/sql/abstraction/dml/Case.java
```diff
@@ -0,0 +1,120 @@
...
+        return caseWhen().when(condition, trueAction);
+    }
+
+    /**
+     * @param condition condition to verify.
+     * @param trueAction action to be executed if the condition is true.
+     * @param falseAction action to be executed if the condition is false.
+     * @return a new "case when" that does the trueAction if the condition is true. Otherwise it runs falseAction.
+     */
+    public static Case caseWhen(final Expression condition, final Expression trueAction,
```
marmelo: Do we need this given that this can be expressed by
```
caseWhen(condition, trueAction).otherwise(falseAction)
```
?

src/main/java/com/feedzai/commons/sql/abstraction/dml/When.java
```diff
@@ -0,0 +1,49 @@
...
+     */
+    public final Expression condition;
+
+    /**
+     * The action to be executed if the condition is true.
+     */
+    public final Expression action;
+
+    /**
+     * @param condition condition to verify.
```
marmelo: Missing Javadoc.

src/main/java/com/feedzai/commons/sql/abstraction/dml/When.java
```diff
@@ -0,0 +1,49 @@
...
+ * reproduced in whole or in part, stored in a retrieval system,
+ * transmitted in any form, or by any means electronic, mechanical,
+ * photocopying, or otherwise, without the prior permission of the owner.
+ *
+ * (c) 2018 Feedzai, Strictly Confidential
+ */
+package com.feedzai.commons.sql.abstraction.dml;
+
+/**
+ * @author Francisco Santos (francisco.santos@feedzai.com)
```
marmelo: Missing Javadoc.

src/main/java/com/feedzai/commons/sql/abstraction/dml/When.java
```diff
@@ -0,0 +1,49 @@
...
+     * @param condition condition to verify.
+     * @param action action to be executed if the condition is true.
+     */
+    private When(final Expression condition, final Expression action) {
+        this.condition = condition;
+        this.action = action;
+    }
+
+    /**
+     * @param condition condition to verify.
```
marmelo: Missing Javadoc.

src/main/java/com/feedzai/commons/sql/abstraction/dml/dialect/SqlBuilder.java
```diff
@@ -15,14 +15,60 @@
  */
 package com.feedzai.commons.sql.abstraction.dml.dialect;
 
-import com.feedzai.commons.sql.abstraction.ddl.*;
-import com.feedzai.commons.sql.abstraction.dml.*;
+import com.feedzai.commons.sql.abstraction.ddl.AlterColumn;
```
marmelo: Please check before/after.  
Not sure if PDB code style uses wildcard imports or not.

src/main/java/com/feedzai/commons/sql/abstraction/dml/dialect/SqlBuilder.java
```diff
@@ -519,6 +565,38 @@ public static View createView(final String name) {
         return new View(name);
     }
 
+    /**
+     * Returns a new "case when" that returns true or false considering the condition.
+     *
+     * @param condition condition to verify.
+     * @return a new "case when" considering the condition.
+     */
+    public static Case caseWhen(final Expression condition) {
```
marmelo: Again, why do we need this?

src/main/java/com/feedzai/commons/sql/abstraction/engine/AbstractTranslator.java
```diff
@@ -331,6 +333,59 @@ public String translate(Update u) {
...
+    }
+
+    /**
+     * Translates a when inside a case expression.
+     *
+     * @param aCase a case expression.
+     * @return a translated when expression.
+     */
+    private StringBuilder caseBuilder(final Case aCase) {
+        List<When> whens = aCase.whens;
```
marmelo: May be `final`.

src/main/java/com/feedzai/commons/sql/abstraction/engine/AbstractTranslator.java
```diff
@@ -331,6 +333,59 @@ public String translate(Update u) {
...
+
+        inject(whens);
+        StringBuilder builder = new StringBuilder();
```
marmelo: May be `final`.

src/main/java/com/feedzai/commons/sql/abstraction/engine/AbstractTranslator.java
```diff
@@ -331,6 +333,59 @@ public String translate(Update u) {
...
+                             elseString);
+    }
+
+    /**
+     * Translates a when inside a case expression.
+     *
+     * @param aCase a case expression.
+     * @return a translated when expression.
+     */
+    private StringBuilder caseBuilder(final Case aCase) {
```
marmelo: I would return a `String`, it would be more clean.  
Your call.

src/main/java/com/feedzai/commons/sql/abstraction/engine/AbstractTranslator.java
```diff
@@ -331,6 +333,59 @@ public String translate(Update u) {
...
+        List<When> whens = aCase.whens;
+
+        inject(whens);
+        StringBuilder builder = new StringBuilder();
+
+        for (int i = 0; i < whens.size(); i++) {
```
marmelo: I believe this would be cleaner if using a stream, something like:

```
whens.stream()
  .peek(this::inject)
  .map(When::translate)
  .collect(Collectors.joining(" "));
```
I did not test it, but it makes conceptually sense.
Maybe after this you don't event need this method and move this to the one above.

That said, functionally it is the same, so it is your call how you design the code.

PR #101, opened 2019-02-11
Support for Cast Operation

### LINKED ISSUE #100: Support for CAST clause.
All SQL engines support CAST clauses.

francisco-polaco:
Support for Cast Operation
This closes #100

### DIFF
--- a/src/main/java/com/feedzai/commons/sql/abstraction/dml/dialect/SqlBuilder.java
+++ b/src/main/java/com/feedzai/commons/sql/abstraction/dml/dialect/SqlBuilder.java
@@ -550,6 +550,17 @@ public final class SqlBuilder {
     }
 
     /**
+     * Creates a cast expression.
+     *
+     * @param expression The expression.
+     * @param type the type to be converted.
+     * @return The cast representation.
+     */
+    public static Cast cast(final Expression expression, final DbColumnType type) {
+        return new Cast(expression, type);
+    }
+
+    /**
      * Returns a new "case when".
      *
      * @return a new "case when".
--- a/src/main/java/com/feedzai/commons/sql/abstraction/dml/dialect/SqlBuilder.java
+++ b/src/main/java/com/feedzai/commons/sql/abstraction/dml/dialect/SqlBuilder.java
@@ -551,6 +551,8 @@ public final class SqlBuilder {
 
     /**
      * Creates a cast expression.
+     * PDB does not support cast to JSON, CLOB and BLOB.
+     * An exception will be thrown when trying to translate the data type.
      *
      * @param expression The expression.
      * @param type the type to be converted.
--- a/src/main/java/com/feedzai/commons/sql/abstraction/dml/result/MySqlResultColumn.java
+++ b/src/main/java/com/feedzai/commons/sql/abstraction/dml/result/MySqlResultColumn.java
@@ -32,4 +32,14 @@ public class MySqlResultColumn extends ResultColumn {
     public MySqlResultColumn(final String name, final Object val) {
         super(name, val);
     }
+
+    @Override
+    public Boolean toBoolean() {
+        try {
+            return super.toBoolean();
+        } catch (ClassCastException e) {
+            // Since booleans are represented using unsigned integers, we need to convert them.
+            return Long.parseLong(val.toString()) != 0;
+        }
+    }
 }
--- a/src/main/java/com/feedzai/commons/sql/abstraction/dml/result/MySqlResultColumn.java
+++ b/src/main/java/com/feedzai/commons/sql/abstraction/dml/result/MySqlResultColumn.java
@@ -38,7 +38,7 @@ public class MySqlResultColumn extends ResultColumn {
         try {
             return super.toBoolean();
         } catch (ClassCastException e) {
-            // Since booleans are represented using unsigned integers, we need to convert them.
+            // Since booleans are represented using unsigned integers, when casting, we need to convert them.
             return Long.parseLong(val.toString()) != 0;
         }
     }
--- a/src/main/java/com/feedzai/commons/sql/abstraction/dml/result/MySqlResultColumn.java
+++ b/src/main/java/com/feedzai/commons/sql/abstraction/dml/result/MySqlResultColumn.java
@@ -33,13 +33,4 @@ public class MySqlResultColumn extends ResultColumn {
         super(name, val);
     }
 
-    @Override
-    public Boolean toBoolean() {
-        try {
-            return super.toBoolean();
-        } catch (ClassCastException e) {
-            // Since booleans are represented using unsigned integers, when casting, we need to convert them.
-            return Long.parseLong(val.toString()) != 0;
-        }
-    }
 }
--- a/src/main/java/com/feedzai/commons/sql/abstraction/dml/result/MySqlResultColumn.java
+++ b/src/main/java/com/feedzai/commons/sql/abstraction/dml/result/MySqlResultColumn.java
@@ -32,5 +32,4 @@ public class MySqlResultColumn extends ResultColumn {
     public MySqlResultColumn(final String name, final Object val) {
         super(name, val);
     }
-
 }
--- a/src/main/java/com/feedzai/commons/sql/abstraction/engine/AbstractTranslator.java
+++ b/src/main/java/com/feedzai/commons/sql/abstraction/engine/AbstractTranslator.java
@@ -21,6 +21,7 @@ import com.feedzai.commons.sql.abstraction.ddl.DropPrimaryKey;
 import com.feedzai.commons.sql.abstraction.ddl.Rename;
 import com.feedzai.commons.sql.abstraction.dml.Between;
 import com.feedzai.commons.sql.abstraction.dml.Case;
+import com.feedzai.commons.sql.abstraction.dml.Cast;
 import com.feedzai.commons.sql.abstraction.dml.Coalesce;
 import com.feedzai.commons.sql.abstraction.dml.Delete;
 import com.feedzai.commons.sql.abstraction.dml.Expression;
@@ -371,6 +372,19 @@ public abstract class AbstractTranslator {
         return String.format("CASE %s %s END",
                              whens,
                              elseString);
+    }
+
+    /**
+     * Translates Cast.
+     *
+     * @param cast a cast expression.
+     * @return cast translation.
+     */
+    public String translate(final Cast cast) {
+        inject(cast.getExpression());
+        return String.format("CAST(%s AS %s)",
+                cast.getExpression().translate(),
+                this.translate(new DbColumn.Builder().type(cast.getType()).build()));
     }
 
     /**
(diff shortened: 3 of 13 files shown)


src/main/java/com/feedzai/commons/sql/abstraction/dml/result/MySqlResultColumn.java
```diff
@@ -32,4 +32,14 @@
     public MySqlResultColumn(final String name, final Object val) {
         super(name, val);
     }
+
+    @Override
+    public Boolean toBoolean() {
+        try {
+            return super.toBoolean();
+        } catch (ClassCastException e) {
+            // Since booleans are represented using unsigned integers, we need to convert them.
```
marmelo: AFAIK the `toBoolean()` is already being called when I read a record from a table with a boolean column.  
Why was this method not required before?

src/main/java/com/feedzai/commons/sql/abstraction/dml/result/MySqlResultColumn.java
```diff
@@ -32,4 +32,14 @@
     public MySqlResultColumn(final String name, final Object val) {
         super(name, val);
     }
+
+    @Override
+    public Boolean toBoolean() {
+        try {
+            return super.toBoolean();
+        } catch (ClassCastException e) {
```
marmelo: Missing `final`.

src/main/java/com/feedzai/commons/sql/abstraction/engine/impl/MySqlTranslator.java
```diff
@@ -275,6 +275,34 @@ public String translate(DbColumn c) {
         }
     }
 
+    @Override
+    public String translateToCast(DbColumn dc) {
```
marmelo: I believe we should have a dedicated method to translate a `DbColumnType` to a string.  
This would be completely independent of the `CAST`.

This way you can have all `impl/*Translator.java`s to define how the types are translated.
And then you would have a very simple `translate(Cast) method.

src/main/java/com/feedzai/commons/sql/abstraction/engine/impl/MySqlTranslator.java
```diff
@@ -300,4 +328,27 @@ public String translateTrue() {
...
+        final Expression expression = cast.getExpression();
+        inject(expression);
+
+        if (cast.getType() == DbColumnType.BOOLEAN) {
+            final DbColumn column = new DbColumn.Builder().type(cast.getType()).build();
+            final String translation = expression.translate()
+                    .replaceAll("'", "")
+                    .toLowerCase();
+
+            if (translation.matches("t|true|1")) {
```
marmelo: Is this "enum" created by you or is this existing logic?

src/test/java/com/feedzai/commons/sql/abstraction/engine/impl/abs/EngineGeneralTest.java
```diff
@@ -1981,6 +1983,39 @@ public void betweenTest() throws DatabaseEngineException {
...
+                .build());
+        engine.persist("TEST", entry().set("COL1", 4).set("COL5", "teste")
+                .build());
+
+        final Query query =
+                select(cast(k("22"), INT).alias("int"),
+                        cast(k(22), STRING).alias("string"),
+                        cast(k("true"), BOOLEAN).alias("bool"),
+                        cast(k("22"), DOUBLE).alias("double"),
+                        cast(k(22), LONG).alias("long"));
```
marmelo: We need to test ALL available `DbColumnType`s.

src/main/java/com/feedzai/commons/sql/abstraction/engine/impl/OracleTranslator.java
```diff
@@ -326,4 +328,27 @@ public String translateTrue() {
...
+        inject(expression);
+
+        if (cast.getType() == DbColumnType.BOOLEAN) {
+            final DbColumn column = new DbColumn.Builder().type(DbColumnType.INT).build();
+            final String translation = expression.translate()
+                    .replaceAll("'", "")
+                    .toLowerCase();
+
+            if (translation.matches("t|true|1")) {
+                return String.format("CAST(1 AS %s)", translateToCast(column));
```
marmelo: What happens if i want to CAST a column instead of a literal?

src/test/java/com/feedzai/commons/sql/abstraction/engine/impl/abs/EngineGeneralTest.java
```diff
@@ -1981,6 +1983,39 @@ public void betweenTest() throws DatabaseEngineException {
         );
     }
 
+    @Test
+    public void testCast() throws DatabaseEngineException {
```
marmelo: You should also test CASTing columns.

src/main/java/com/feedzai/commons/sql/abstraction/dml/dialect/SqlBuilder.java
```diff
@@ -549,6 +549,17 @@ public static View createView(final String name) {
         return new View(name);
     }
 
+    /**
+     * Creates a cast expression.
+     *
```
marmelo: Add something like

```
        // PDB does not support cast to JSON, CLOB and BLOB.
        // An exception will be thrown when trying to translate the data type.
```

src/main/java/com/feedzai/commons/sql/abstraction/dml/result/MySqlResultColumn.java
```diff
@@ -32,4 +32,5 @@
     public MySqlResultColumn(final String name, final Object val) {
         super(name, val);
     }
+
```
marmelo: :disappointed:

src/main/java/com/feedzai/commons/sql/abstraction/engine/impl/H2Translator.java
```diff
@@ -292,6 +293,29 @@ public String translate(DbColumn c) {
         }
     }
 
+    @Override
+    public String translate(final DbColumnType type) {
+        switch (type) {
+            case BOOLEAN:
+                return "BOOLEAN";
+
```
marmelo: I don't like to put newlines in cases.  
Just a comment, do whatever you want. :)

src/main/java/com/feedzai/commons/sql/abstraction/engine/AbstractTranslator.java
```diff
@@ -466,6 +510,14 @@ public String translate(final Case aCase) {
      */
     public abstract String translate(DbColumn dc);
 
+    /**
+     * Translates {@link DbColumnType}.
+     *
+     * @param type The object to translate.
+     * @return The string representation of the given object.
+     */
+    public abstract String translate(DbColumnType type);
```
marmelo: This was not what I had in my mind.  
I don't believe this translate method makes sense.

I was expecting that the contents of this method was inside each `translate(Cast)` directly.  
I believe this is better because it may be misleading because the `translate(DbColumn)` has some specific types of the CAST operation.

What do you think?

src/main/java/com/feedzai/commons/sql/abstraction/engine/impl/OracleTranslator.java
```diff
@@ -294,6 +297,29 @@ public String translate(DbColumn c) {
         }
     }
 
+    @Override
+    public String translate(final DbColumnType type) {
+        switch (type) {
+            case BOOLEAN:
+                return "NUMBER(1,0)";
```
marmelo: We need to keep the types as compatible as possible with the `translate(DbColumn c)` because we will have PDB trying to interpret them.

That said, shouldn't this be `char`?

src/main/java/com/feedzai/commons/sql/abstraction/engine/impl/OracleTranslator.java
```diff
@@ -294,6 +297,29 @@ public String translate(DbColumn c) {
...
+                return "BINARY_DOUBLE";
+
+            case INT:
+                return "INT";
+
+            case LONG:
+                return "NUMBER(19,0)";
+
+            case STRING:
+                return format("VARCHAR(%s)", properties.getProperty(VARCHAR_SIZE));
```
marmelo: We should not set the size. Can we just use `VARCHAR`?

src/main/java/com/feedzai/commons/sql/abstraction/engine/AbstractTranslator.java
```diff
@@ -373,6 +375,48 @@ public String translate(final Case aCase) {
...
+     * If it can't, it returns null.
+     *
+     * @param expression the expression to translate the boolean value.
+     * @return the boolean value if possible, null otherwise.
+     */
+    protected String tryBooleanTranslate(final Expression expression) {
+
+        // Have the translation easier to match the regex below.
+        final String translation = expression.translate()
+                .replaceAll("'", "")
```
marmelo: I am not comfortable with this. Need to think.

src/main/java/com/feedzai/commons/sql/abstraction/engine/AbstractTranslator.java
```diff
@@ -373,6 +375,48 @@ public String translate(final Case aCase) {
...
+                .toLowerCase();
+
+        // If matches a true value, then set the value to '1'.
+        // If matches a false value, then set the value to '0'.
+        // Else, just let it be.
+        if (translation.matches("t|true|1")) {
```
marmelo: I am not comfortable with this. Need to think.

src/main/java/com/feedzai/commons/sql/abstraction/engine/AbstractTranslator.java
francisco-polaco: I understand your point. It makes sense.
I'm not sure if we should make that switch-case method a private one, for each class, or have the code in the cast translate itself.
The first allows for shorter methods, but I believe it will impact readability because the exception is not visible.

What do you think?

src/main/java/com/feedzai/commons/sql/abstraction/engine/impl/OracleTranslator.java
francisco-polaco: I think it's debatable. For what I saw in the community, there's not a consensus about it. Some people prefer a char, like 't'/'f', others believe that a "BIT" it's better.

I honestly don't like representing things with text, unless I have to. Why would it be better to be a `char`?

src/main/java/com/feedzai/commons/sql/abstraction/engine/impl/OracleTranslator.java
francisco-polaco: Nop. In Oracle and DB2 you are forced to specify a size.

src/main/java/com/feedzai/commons/sql/abstraction/engine/AbstractTranslator.java
francisco-polaco: We should discuss this offline.

src/main/java/com/feedzai/commons/sql/abstraction/engine/AbstractTranslator.java
francisco-polaco: We should discuss this offline.

PR #103, opened 2019-02-11
Support for With clause.

### LINKED ISSUE #102: Support for WITH clause.
Supported by all engines.

francisco-polaco:
This closes #102

### DIFF
--- a/src/main/java/com/feedzai/commons/sql/abstraction/dml/With.java
+++ b/src/main/java/com/feedzai/commons/sql/abstraction/dml/With.java
@@ -1,0 +1,63 @@
+/*
+ * Copyright 2018 Feedzai
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.feedzai.commons.sql.abstraction.dml;
+
+import com.google.common.collect.Lists;
+import org.apache.commons.lang3.tuple.ImmutablePair;
+
+import java.util.List;
+
+/**
+ * With SQL Expression.
+ *
+ * @author Francisco Santos (francisco.santos@feedzai.com)
+ * @since 2.3.1
+ */
+public class With extends Expression {
+
+    private final List<ImmutablePair<Name, Expression>> withs;
+
+    private Expression then;
+
+    public With(final String name, final Expression expression) {
+        this.withs = Lists.newArrayList(new ImmutablePair<>(new Name(name), expression));
+    }
+
+    public List<ImmutablePair<Name, Expression>> getWiths() {
+        return withs;
+    }
+
+    public Expression getThen() {
+        return then;
+    }
+
+    public With andWith(final String name, final Expression expression) {
+        this.withs.add(new ImmutablePair<>(new Name(name), expression));
+        return this;
+    }
+
+    public With then(final Expression thenExpression) {
+        this.then = thenExpression;
+        return this;
+    }
+
+    @Override
+    public String translate() {
+        final String translate = translator.translate(this);
+        System.out.println("lol "+ translate);
+        return translate;
+    }
+}
--- a/src/main/java/com/feedzai/commons/sql/abstraction/dml/With.java
+++ b/src/main/java/com/feedzai/commons/sql/abstraction/dml/With.java
@@ -24,7 +24,7 @@ import java.util.List;
  * With SQL Expression.
  *
  * @author Francisco Santos (francisco.santos@feedzai.com)
- * @since 2.3.1
+ * @since 2.4.0
  */
 public class With extends Expression {
 
--- a/src/main/java/com/feedzai/commons/sql/abstraction/dml/With.java
+++ b/src/main/java/com/feedzai/commons/sql/abstraction/dml/With.java
@@ -28,27 +28,62 @@ import java.util.List;
  */
 public class With extends Expression {
 
-    private final List<ImmutablePair<Name, Expression>> withs;
+    /**
+     * The clauses.
+     */
+    private final List<ImmutablePair<Name, Expression>> clauses;
 
+    /**
+     * Expression to execute given the with clauses.
+     */
     private Expression then;
 
+    /**
+     * Creates a new WITH expression.
+     *
+     * @param alias expression alias.
+     * @param expression expression to be aliased.
+     */
     public With(final String alias, final Expression expression) {
-        this.withs = Lists.newArrayList(new ImmutablePair<>(new Name(alias), expression));
+        this.clauses = Lists.newArrayList(new ImmutablePair<>(new Name(alias), expression));
     }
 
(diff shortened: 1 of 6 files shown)


src/main/java/com/feedzai/commons/sql/abstraction/dml/Union.java
```diff
@@ -0,0 +1,93 @@
...
+import java.util.LinkedList;
+import java.util.List;
+
+/**
+ * The union clause.
+ *
+ * @author Francisco Santos (francisco.santos@feedzai.com)
+ * @since 2.3.1
+ */
+public class Union extends Expression {
```
marmelo: With or Union?  
You should have an independent pull request for Union before implementing With.

src/main/java/com/feedzai/commons/sql/abstraction/engine/AbstractTranslator.java
```diff
@@ -335,6 +339,29 @@ public String translate(Update u) {
...
+            injector.injectMembers(aWith.getRight());
+        });
+
+        final String withStatements = withs.stream()
+                .map(aWith -> aWith.getLeft().translate() + " AS (" + aWith.getRight().translate() + ")")
+                .collect(Collectors.joining(", "));
+
+        final Expression then = with.getThen();
+
+        if (then != null) {
```
marmelo: Does it makes sense to have a WITH without the "then"?  
What is an example usage of that?

src/main/java/com/feedzai/commons/sql/abstraction/engine/impl/MySqlTranslator.java
```diff
@@ -275,6 +276,11 @@ public String translate(DbColumn c) {
         }
     }
 
+    @Override
+    public String translate(final With with) {
+        throw new OperationNotSupportedRuntimeException("MySQL 5.7 does not support WITH");
```
marmelo: Please omit the version, you cannot know which version is being used.  
Although wrong, lets just assume that all MySQL versions do not support this.

src/main/java/com/feedzai/commons/sql/abstraction/engine/AbstractTranslator.java
francisco-polaco: It does not make sense. However, enforcing the declaration of `then` statement, before all the "withs" are declared, seems strange.
`with("a", <query>, <then-query>);`
What do you think?

src/main/java/com/feedzai/commons/sql/abstraction/dml/Union.java
francisco-polaco: There was a problem with the rebase. Going to fix it.

src/main/java/com/feedzai/commons/sql/abstraction/engine/impl/MySqlTranslator.java
francisco-polaco: Ok

src/main/java/com/feedzai/commons/sql/abstraction/dml/With.java
```diff
@@ -0,0 +1,63 @@
...
+
+/**
+ * With SQL Expression.
+ *
+ * @author Francisco Santos (francisco.santos@feedzai.com)
+ * @since 2.3.1
+ */
+public class With extends Expression {
+
+    private final List<ImmutablePair<Name, Expression>> withs;
```
marmelo: Variables / Methods / Constructor Javadoc is missing.

src/main/java/com/feedzai/commons/sql/abstraction/dml/dialect/SqlBuilder.java
```diff
@@ -498,6 +498,17 @@ public static StringAgg stringAgg(final Expression column) {
...
     }
 
+    /**
+     * The With clause.
+     *
+     * @param name The name to be assigned to the expression.
+     * @param expression The expression.
+     * @return The With clause.
+     */
+    public static With with(final String name, final Expression expression) {
```
marmelo: Should `name` be called `alias` instead?

src/main/java/com/feedzai/commons/sql/abstraction/engine/AbstractTranslator.java
marmelo: Please keep it the way it is.

src/main/java/com/feedzai/commons/sql/abstraction/engine/impl/MySqlTranslator.java
```diff
@@ -275,6 +276,11 @@ public String translate(DbColumn c) {
         }
     }
 
+    @Override
+    public String translate(final With with) {
+        throw new OperationNotSupportedRuntimeException("MySQL does not support WITH");
```
marmelo: Missing `.` at the end of phrase.

src/test/java/com/feedzai/commons/sql/abstraction/engine/impl/abs/EngineGeneralTest.java
```diff
@@ -1989,6 +1992,109 @@ public void betweenTest() throws DatabaseEngineException {
         );
     }
 
+    @Test
+    public void testWith() throws DatabaseEngineException {
+        if (this.engine instanceof MySqlEngine) {
```
marmelo: This is not cool given that this is an abstract test.  
I thought this class was abstract and each engine would extend it but I have just realize that this is not true.

That said, I suggest not comparing `this.engine` but `config.engine` because it is a string an you may remove the dependency.
But again, this is better but still not cool. :/

src/test/java/com/feedzai/commons/sql/abstraction/engine/impl/abs/EngineGeneralTest.java
```diff
@@ -1989,6 +1992,109 @@ public void betweenTest() throws DatabaseEngineException {
...
+                .build());
+        engine.persist("TEST", entry().set("COL1", 2).set("COL5", "ana")
+                .build());
+        engine.persist("TEST", entry().set("COL1", 3).set("COL5", "rita")
+                .build());
+        engine.persist("TEST", entry().set("COL1", 4).set("COL5", "rui")
+                .build());
+
+        final With with = with("friends", select(all())
+                                                .from(table("TEST")))
```
marmelo: Review indentation.

src/test/java/com/feedzai/commons/sql/abstraction/engine/impl/abs/EngineGeneralTest.java
```diff
@@ -1989,6 +1992,109 @@ public void betweenTest() throws DatabaseEngineException {
         );
     }
 
+    @Test
+    public void testWith() throws DatabaseEngineException {
+        if (this.engine instanceof MySqlEngine) {
+            // MySQL does not support With
+            return;
```
marmelo: Should we assert that if MySQL you get the not supported exception?

marmelo: I believe this makes sense, you should assert different things according to the engine.

src/main/java/com/feedzai/commons/sql/abstraction/dml/dialect/SqlBuilder.java
francisco-polaco: Yes

marmelo: Yes, but... you did not change. :)

francisco-polaco: Yup, I forgot to change on the SQLBuilder. Done :)

PR #109, opened 2019-02-28
Enclose of Values, With, Union and Cast operations.

francisco-polaco:
These operations were not considering its enclose property.


marmelo:
@francisco-polaco no need to unit test this?

PR #111, opened 2019-03-18
Issue#110 - Large Values cause StackOverflow when using Unions

### LINKED ISSUE #110: Large Values cause StackOverflow when using Unions
Idea:
 - Apply binary tree of unions.

francisco-polaco:
Applied a binary tree of unions.
This closes #110

### DIFF
--- a/src/main/java/com/feedzai/commons/sql/abstraction/engine/AbstractTranslator.java
+++ b/src/main/java/com/feedzai/commons/sql/abstraction/engine/AbstractTranslator.java
@@ -500,33 +500,23 @@ public abstract class AbstractTranslator {
      * @return the resulting union
      */
     protected Union rowsToUnion(final List<Expression> rows) {
-        return toUnionTree(rows);
-    }
-
-    /**
-     * Create an union in form of a binary tree from a list of rows.
-     * The tree shape is to prevent stack overflow on some
-     * database engines when the list of rows is too big.
-     *
-     * @param rows the list of rows.
-     * @return an union in form of a binary tree.
-     */
-    private Union toUnionTree(final List<Expression> rows) {
+        // Create an union in form of a binary tree from a list of rows.
+        // The tree shape is to prevent stack overflow on some
+        // database engines when the list of rows is too big.
 
         List<Expression> rowsWithSelect = new ArrayList<>(rows);
-        
+
         while (rowsWithSelect.size() > 2) {
             final List<Expression> newRowsWithSelect = new ArrayList<>();
 
             // Put the first and second rowsWithSelect together forming an UNION ALL.
             // Then the third and forth, and so on.
-            for (int i = 1; i < rowsWithSelect.size(); i++) {
-                if ((i - 1) % 2 == 0) {
-                    final Expression left = rowsWithSelect.get(i - 1);
-                    final Expression right = rowsWithSelect.get(i);
+            // To do this, we move though the rows with a delta of 2, to group the rows in pairs.
+            for (int i = 1; i < rowsWithSelect.size(); i+=2) {
+                final Expression left = rowsWithSelect.get(i - 1);
+                final Expression right = rowsWithSelect.get(i);
 
-                    newRowsWithSelect.add(union(left, right).all().enclose());
-                }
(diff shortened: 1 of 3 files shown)


src/main/java/com/feedzai/commons/sql/abstraction/engine/AbstractTranslator.java
```diff
@@ -492,6 +493,54 @@ public String translate(final Values.Row row) {
...
+
+    /**
+     * Create an union in form of a binary tree from a list of rows.
+     * The tree shape is to prevent stack overflow on some
+     * database engines when the list of rows is too big.
+     *
+     * @param rows the list of rows.
+     * @return an union in form of a binary tree.
+     */
+    private Union toUnionTree(final List<Expression> rows) {
```
marmelo: Why create a new method?  
Shouldn't this be inside `rowsToUnion`?

src/main/java/com/feedzai/commons/sql/abstraction/engine/AbstractTranslator.java
```diff
@@ -492,6 +493,54 @@ public String translate(final Values.Row row) {
...
+
+        List<Expression> rowsWithSelect = new ArrayList<>(rows);
+        
+        while (rowsWithSelect.size() > 2) {
```
marmelo: Interesting. My first approach would be to use recursion.  
But AFAIK this implementation is is also O(n log n), let me check.

src/main/java/com/feedzai/commons/sql/abstraction/engine/AbstractTranslator.java
```diff
@@ -492,6 +493,54 @@ public String translate(final Values.Row row) {
...
+            final List<Expression> newRowsWithSelect = new ArrayList<>();
+
+            // Put the first and second rowsWithSelect together forming an UNION ALL.
+            // Then the third and forth, and so on.
+            for (int i = 1; i < rowsWithSelect.size(); i++) {
+                if ((i - 1) % 2 == 0) {
```
marmelo: Why do this instead of `i+=2` in the above line?  
Am I missing something?

src/main/java/com/feedzai/commons/sql/abstraction/engine/impl/MySqlTranslator.java
```diff
@@ -286,6 +287,14 @@ public String translate(final With with) {
         throw new OperationNotSupportedRuntimeException("MySQL does not support WITH.");
     }
 
+    @Override
+    protected Union rowsToUnion(final List<Expression> rows) {
+        // By default, use UNION ALL to express VALUES.
+        // MySQL does not support modeling UNION using binary tree due to a bug prior to 8.0
+        // https://bugs.mysql.com/bug.php?id=25734
+        return union(rows).all();
```
marmelo: Does MySQL break with lots of values?

francisco-polaco: MySQL has a bug regarding parenthesis parsing when in the presence of union.
It was fixed in the last release but PDB does not support it.

src/main/java/com/feedzai/commons/sql/abstraction/engine/AbstractTranslator.java
francisco-polaco: You are right, it does not make sense :p 
When I wrote this I wanted to make sure people understood that the binary tree was there for a reason. But a comment is enough.

src/main/java/com/feedzai/commons/sql/abstraction/engine/AbstractTranslator.java
francisco-polaco: ;) You're right

src/main/java/com/feedzai/commons/sql/abstraction/engine/AbstractTranslator.java
francisco-polaco: I usually only write something recursive when it's supernatural. Otherwise, I stick with an iterative approach.

PR #115, opened 2019-05-07
[Oracle] Batch inserts/updates with LOBs don't free temporary LOB (fixes #114)

### LINKED ISSUE #114: [Oracle] Batch inserts/updates with LOBs don't free temporary LOB
On batch updates with LOB columns with a size greater than 4000 bytes, the oracle driver uses a temporary LOB that is never freed because PDB uses a single PreparedStatement for the whole session.

This can be reproduced by doing multiple inserts with the batch update interface and checking that the temporary LOB space used by the connection keeps growing -- use select CACHE_LOBS
from V$TEMPORARY_LOBS to verify this.

leitaop:
On batch updates with LOB columns with a size greater than 4000 bytes, the oracle driver uses a temporary LOB that is never freed because PDB uses a single PreparedStatement for the whole session. This commit avoids this resource leak by using explicit LOBs on these circumstances, and releasing them explicitly when the batch is flushed. The OracleEngine was modified by adding an attribute with the actions that must be executed after the flush of the current batch, and ensuring that actions are added on addBatch() and performed on flush(). The state of this attribute reflects what was added to the batch of the current insert PreparedStatement, so it has the same resource release requirements as that PreparedStatement.

### DIFF
--- a/src/main/java/com/feedzai/commons/sql/abstraction/engine/impl/OracleEngine.java
+++ b/src/main/java/com/feedzai/commons/sql/abstraction/engine/impl/OracleEngine.java
@@ -233,7 +233,7 @@ public class OracleEngine extends AbstractDatabaseEngine {
                             // it when the PreparedStatement is closed. This will be closed
                             // explicitly when the batch is flushed.
                             final Blob blob = ps.getConnection().createBlob();
-                            blob.setBytes(1, objectToArray(val));
+                            blob.setBytes(1, valArray);
                             ps.setBlob(i, blob);
                             postFlushActions.add(blob::free);
                         } else {
@@ -254,7 +254,7 @@ public class OracleEngine extends AbstractDatabaseEngine {
                                 // it when the PreparedStatement is closed. This will be closed
                                 // explicitly when the batch is flushed.
                                 final Clob clob = ps.getConnection().createClob();
-                                clob.setString(1, (String) val);
+                                clob.setString(1, valString);
                                 ps.setClob(i, clob);
                                 postFlushActions.add(clob::free);
                             } else {
--- a/src/main/java/com/feedzai/commons/sql/abstraction/engine/impl/OracleEngine.java
+++ b/src/main/java/com/feedzai/commons/sql/abstraction/engine/impl/OracleEngine.java
@@ -38,6 +38,8 @@ import com.google.common.collect.ImmutableList;
 import oracle.jdbc.OraclePreparedStatement;
 import oracle.jdbc.OracleTypes;
 
+import java.sql.Blob;
+import java.sql.Clob;
 import java.sql.Connection;
 import java.sql.PreparedStatement;
 import java.sql.ResultSet;
@@ -45,9 +47,11 @@ import java.sql.SQLException;
 import java.sql.Statement;
 import java.sql.Types;
 import java.util.ArrayList;
+import java.util.Deque;
 import java.util.LinkedHashMap;
 import java.util.List;
 import java.util.Map;
+import java.util.concurrent.ConcurrentLinkedDeque;
 
 import static com.feedzai.commons.sql.abstraction.util.StringUtils.md5;
 import static com.feedzai.commons.sql.abstraction.util.StringUtils.quotize;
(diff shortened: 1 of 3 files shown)


src/main/java/com/feedzai/commons/sql/abstraction/engine/impl/OracleEngine.java
```diff
@@ -162,17 +226,40 @@ protected int entityToPreparedStatement(final DbEntity entity, final PreparedSta
...
                     case BLOB:
-                        ps.setBytesForBlob(i, objectToArray(val));
+                        final byte[] valArray = objectToArray(val);
+                        if (fromBatch && valArray.length > MIN_SIZE_FOR_BLOB) {
+                            // Use a blob for columns greater than 4K when inside a batch
+                            // update because the Oracle driver creates one and only releases
+                            // it when the PreparedStatement is closed. This will be closed
+                            // explicitly when the batch is flushed.
+                            final Blob blob = ps.getConnection().createBlob();
+                            blob.setBytes(1, objectToArray(val));
```
jmf-tls:
```suggestion
                            blob.setBytes(1, valArray);
```
`objectToArray` is going to serialize _val_ again, but we already have the result of that conversion

src/main/java/com/feedzai/commons/sql/abstraction/engine/impl/OracleEngine.java
```diff
@@ -162,17 +226,40 @@ protected int entityToPreparedStatement(final DbEntity entity, final PreparedSta
...
                         if (val instanceof String) {
-                            ps.setStringForClob(i, (String) val);
+                            final String valString = (String) val;
+                            if (fromBatch && valString.length() > MIN_SIZE_FOR_CLOB) {
+                                // Use a clob for columns greater than 35K when inside a batch
+                                // update because the Oracle driver creates one and only releases
+                                // it when the PreparedStatement is closed. This will be closed
+                                // explicitly when the batch is flushed.
+                                final Clob clob = ps.getConnection().createClob();
+                                clob.setString(1, (String) val);
```
jmf-tls:
```suggestion
                                clob.setString(1, valString);
```
same as above (even though in this case there is no performance difference)

src/test/java/com/feedzai/commons/sql/abstraction/engine/impl/oracle/OracleEngineSchemaTest.java
```diff
@@ -258,6 +264,85 @@ public void testBlobWithGreaterSizes() throws Exception {
         }
     }
 
+    /**
+     * This is a regression test for https://github.com/feedzai/pdb/issues/114. It inserts
+     * 10 rows containing a BLOB column using the batch update interface and ensures that
+     * no temp lOBs are left.
```
jmf-tls:
```suggestion
     * no temp LOBs are left.
```
nitpicking :)

src/test/java/com/feedzai/commons/sql/abstraction/engine/impl/oracle/OracleEngineSchemaTest.java
```diff
@@ -258,6 +264,85 @@ public void testBlobWithGreaterSizes() throws Exception {
...
+     */
+    @Test
+    public void testBlobBatchInsertClearsResources() throws Exception {
+        testLobBatchInsertClearsResources(BLOB);
+    }
+
+    /**
+     * This is a regression test for https://github.com/feedzai/pdb/issues/114. It inserts
+     * 10 rows containing a CLOB column using the batch update interface and ensures that
+     * no temp lOBs are left.
```
jmf-tls:
```suggestion
     * no temp LOBs are left.
```
(same as above)

src/test/java/com/feedzai/commons/sql/abstraction/engine/impl/oracle/OracleEngineSchemaTest.java
```diff
@@ -258,6 +264,85 @@ public void testBlobWithGreaterSizes() throws Exception {
...
+     * @since 2.4.2
+     */
+    @Test
+    public void testClobBatchInsertClearsResources() throws Exception {
+        testLobBatchInsertClearsResources(CLOB);
+    }
+
+    /**
+     * This is a regression test for https://github.com/feedzai/pdb/issues/114. It inserts
+     * 10 rows using the batch update interface and ensures that no temp lOBs are left.
```
jmf-tls:
```suggestion
     * 10 rows using the batch update interface and ensures that no temp LOBs are left.
```
(same as above)

jmf-tls:
Don't do it like me: instead of using simply "(#114)" in the commit message,
I'd suggest you use "(fixes #114)", which auto-closes the issue

src/test/java/com/feedzai/commons/sql/abstraction/engine/impl/oracle/OracleEngineSchemaTest.java
```diff
@@ -258,6 +264,75 @@ public void testBlobWithGreaterSizes() throws Exception {
...
 
+    /**
+     * This is a regression test for https://github.com/feedzai/pdb/issues/114. It inserts
+     * 10 rows using the batch update interface and ensures that no temp LOBs are left.
+     *
+     * @throws Exception Should not be thrown.
+     * @since 2.4.2
+     */
+    @Test
+    public void testLobBatchInsertClearsLobResources() throws Exception {
```
jmf-tls: Good idea to merge the CLOB and BLOB testing into a single test

PR #119, opened 2019-07-05
[Oracle] Also clean temporary segment for LOB data when LOBs are freed (fixes #114)

### LINKED ISSUE #114: [Oracle] Batch inserts/updates with LOBs don't free temporary LOB

jmf-tls:
Summary:
The changes in #115 made sure that the LOBs were freed after use, which
 cleared them from LOB cache.
However, this cache is merely a locator for temporary LOB data; to
 really remove that data after usage, the session needs to enable the
 Oracle 60025 Event for temp segment cleanup used for temp lobs.

For more information on the subject see
http://ksun-oracle.blogspot.com/2015/06/temporary-lob-memory-allocation-and.html

### DIFF
--- a/src/main/java/com/feedzai/commons/sql/abstraction/engine/configuration/PdbProperties.java
+++ b/src/main/java/com/feedzai/commons/sql/abstraction/engine/configuration/PdbProperties.java
@@ -27,6 +27,7 @@ import java.util.Properties;
 
 import static com.feedzai.commons.sql.abstraction.util.Constants.DEFAULT_ALLOW_COLUMN_DROP;
 import static com.feedzai.commons.sql.abstraction.util.Constants.DEFAULT_BLOB_BUFFER_SIZE;
+import static com.feedzai.commons.sql.abstraction.util.Constants.DEFAULT_DISABLE_LOB_CACHING;
 import static com.feedzai.commons.sql.abstraction.util.Constants.DEFAULT_FETCH_SIZE;
 import static com.feedzai.commons.sql.abstraction.util.Constants.DEFAULT_ISOLATION_LEVEL;
 import static com.feedzai.commons.sql.abstraction.util.Constants.DEFAULT_MAXIMUM_TIME_BATCH_SHUTDOWN;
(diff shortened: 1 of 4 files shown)


leitaop:
LGTM, seems to solve the problem completely. 

I just have one concern from reading the oracle documentation:
https://docs.oracle.com/cd/B28359_01/server.111/b28278/e60000.htm#ORA-60001

Maybe we should have this configurable.

Quote from the link above:
"ORA-60025: Event for temp segment cleanup used for temp lobs
Cause: Temp LOB segments used for temporary LOBs are deleted only on session exit which may lead to large amounts of memory being held across multiple sessions.

Action: Setting this event will cause temporary LOB segments to be freed when there are no active temporary LOBs in the session. Setting this event will have a significant performance impact as it can cause temporary lob segments to be allocated and deleted many times during a session rather than once per session. Use this event only when temporary LOB segment memory use is an issue."

jmf-tls:
@leitaop 
As discussed, I suspected that this might have a performance impact but didn't find documentation, so thanks for the link!
The cleaning of LOB data only occurs if the LOBs have already been freed from the LOB cache (where the pointers/locators are stored); nevertheless I suppose there may be still some overhead in handling the trace data anyway, so I made this behavior configurable (with the default being disabled, and working as before).

I also got confirmation that the problem is now solved when the temp segment cleanup is enabled, along with the LOBs being freed.

src/main/java/com/feedzai/commons/sql/abstraction/engine/configuration/PdbProperties.java
```diff
@@ -144,6 +145,11 @@
      * Property that indicates the lobs should be compressed. This depends on the database implementation.
      */
     public static final String COMPRESS_LOBS = "pdb.compress_lobs";
+    /**
+     * Property that indicates if LOB data caching should be disabled, to avoid consuming too much memory and/or disk
+     * space in the DB server. This depends on the database implementation.
+     */
```
leitaop: Suggest you mention this is applicable only to Oracle.

leitaop:
Just made a minor comment, LGTM and is ready to land.

src/main/java/com/feedzai/commons/sql/abstraction/engine/configuration/PdbProperties.java
jmf-tls: I see what you mean...
I added the "This depends on the database implementation" note, just like the _compress_lobs_ property has.

Both properties currently do something only on Oracle, but if an issue with LOB data increasing disk/memory usage is found in other DB implementations, this property is generic enough to also apply as a toggle for custom solutions specific for those implementations.

PR #120, opened 2019-07-22
[PostgreSQL] Keep old behavior for default SSL mode

jmf-tls:
Summary:
The PostgreSQL driver was updated to the latest version 42.2.6 in commit
 3dc357883b819c7c6bab27e122e1598ed17450f6; since 42.2.5, when SSL is
 enabled but SSL mode is not set, it is assumed to be "verify-full" by
 default.

 SSL is considered enabled if the "ssl" option present in JDBC URL,
  either without arguments or with argument "true".

In the previous driver versions, if SSL was enabled and SSL mode was not
 set, then "require" would be the default.

To keep compatibility with the old behavior and avoid requiring
 configuration changes, this commit now changes the JDBC URL internally
 before connecting by setting "sslmode" to "require" if it is not set
 but SSL is enabled.

The SSL mode configuration should be explicitly set, and this code may
 be reverted in the future.

### DIFF
--- a/src/main/java/com/feedzai/commons/sql/abstraction/engine/impl/PostgreSqlEngine.java
+++ b/src/main/java/com/feedzai/commons/sql/abstraction/engine/impl/PostgreSqlEngine.java
@@ -119,6 +119,7 @@ public class PostgreSqlEngine extends AbstractDatabaseEngine {
         final boolean sslEnabled = PGProperty.SSL.getBoolean(parsedProps) || "".equals(PGProperty.SSL.get(parsedProps));
         final String sslmode = PGProperty.SSL_MODE.get(parsedProps);
         if (sslEnabled && sslmode == null) {
+            logger.trace("SSL enabled without SSL mode specified: \"require\" will be used by default, appended to JDBC URL");
             return jdbc.concat(LEGACY_DEFAULT_SSL_MODE);
         }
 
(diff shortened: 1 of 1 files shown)


src/main/java/com/feedzai/commons/sql/abstraction/engine/impl/PostgreSqlEngine.java
```diff
@@ -90,6 +106,25 @@ public PostgreSqlEngine(PdbProperties properties) throws DatabaseEngineException
...
+            return jdbc;
+        }
+
+        // If SSL is enabled ("ssl" option is present in JDBC URL, with argument "true" or without arguments)
+        // but SSL mode ("sslmode" option) is not set, use legacy behavior - consider "sslmode" = "require"
+        // NOTE: the properties should be explicitly set, this code may be reverted in the future
+        final boolean sslEnabled = PGProperty.SSL.getBoolean(parsedProps) || "".equals(PGProperty.SSL.get(parsedProps));
+        final String sslmode = PGProperty.SSL_MODE.get(parsedProps);
+        if (sslEnabled && sslmode == null) {
+            return jdbc.concat(LEGACY_DEFAULT_SSL_MODE);
```
joaonuno: Should we log this?

jmf-tls: seems fair, I'll add log TRACE

PR #121, opened 2019-08-22
Add socket connection timeout.

TiagoMSSantos:
The pdb doesn't have a connection timeout, so when the connection to the database fails abruptly, the threads block waiting for a response of a submitted query.
This new branch adds a connection and a login connection timeout (defined by the user) so it doesn't block the system.

### DIFF
--- a/src/main/java/com/feedzai/commons/sql/abstraction/engine/AbstractDatabaseEngine.java
+++ b/src/main/java/com/feedzai/commons/sql/abstraction/engine/AbstractDatabaseEngine.java
@@ -38,6 +38,7 @@ import com.feedzai.commons.sql.abstraction.util.PreparedStatementCapsule;
 import com.google.common.collect.ImmutableList;
 import com.google.inject.Inject;
 import com.google.inject.Injector;
+import oracle.jdbc.driver.OracleConnection;
 import org.apache.commons.lang3.StringUtils;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
@@ -222,6 +223,15 @@ public abstract class AbstractDatabaseEngine implements DatabaseEngine {
     }
 
     /**
+     * Get the properties to be passed to the database connection that will be created. Since these
+     * properties are specific to the database engine, it is up to the {@link AbstractDatabaseEngine}
+     * implementation to specify them. This is used, for instance, to set the database connection timeout.
+     *
+     * @return the properties of the connection to the database.
+     */
+    protected abstract Properties getDBProperties();
+
+    /**
      * Connects to the database.
      *
      * @throws Exception If connection is not possible, or failed to decrypt username/password if encryption was provided.
@@ -252,10 +262,15 @@ public abstract class AbstractDatabaseEngine implements DatabaseEngine {
 
         final String jdbc = getFinalJdbcConnection(this.properties.getJdbc());
 
-        this.conn = DriverManager.getConnection(
-                jdbc,
-                username,
-                password);
+        final Properties props = getDBProperties();
+
+        // Need to check if not null, because setProperty method doesn't accept null values
+        if (username != null && password != null) {
+            props.setProperty("user", username);
+            props.setProperty("password", password);
+        }
+
+        this.conn = DriverManager.getConnection(jdbc, props);
 
         if (this.properties.isSchemaSet()) {
             setSchema(this.properties.getSchema());
--- a/src/main/java/com/feedzai/commons/sql/abstraction/engine/configuration/PdbProperties.java
+++ b/src/main/java/com/feedzai/commons/sql/abstraction/engine/configuration/PdbProperties.java
@@ -30,6 +30,7 @@ import static com.feedzai.commons.sql.abstraction.util.Constants.DEFAULT_BLOB_BU
 import static com.feedzai.commons.sql.abstraction.util.Constants.DEFAULT_DISABLE_LOB_CACHING;
 import static com.feedzai.commons.sql.abstraction.util.Constants.DEFAULT_FETCH_SIZE;
 import static com.feedzai.commons.sql.abstraction.util.Constants.DEFAULT_ISOLATION_LEVEL;
+import static com.feedzai.commons.sql.abstraction.util.Constants.DEFAULT_LOGIN_TIMEOUT;
 import static com.feedzai.commons.sql.abstraction.util.Constants.DEFAULT_MAXIMUM_TIME_BATCH_SHUTDOWN;
 import static com.feedzai.commons.sql.abstraction.util.Constants.DEFAULT_MAX_IDENTIFIER_SIZE;
 import static com.feedzai.commons.sql.abstraction.util.Constants.DEFAULT_RECONNECT_ON_LOST;
@@ -37,6 +38,7 @@ import static com.feedzai.commons.sql.abstraction.util.Constants.DEFAULT_RETRY_I
 import static com.feedzai.commons.sql.abstraction.util.Constants.DEFAULT_SCHEMA_POLICY;
 import static com.feedzai.commons.sql.abstraction.util.Constants.DEFAULT_SECRET_LOCATION;
 import static com.feedzai.commons.sql.abstraction.util.Constants.DEFAULT_COMPRESS_LOBS;
+import static com.feedzai.commons.sql.abstraction.util.Constants.DEFAULT_SOCKET_TIMEOUT;
 import static com.feedzai.commons.sql.abstraction.util.Constants.DEFAULT_VARCHAR_SIZE;
 
 /**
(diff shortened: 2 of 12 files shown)


TiagoMSSantos:
I just want to add that this fix doesn't work with Oracle databases, so don't focus too much on the code, but rather on the approach of adding new settings to configure the socket connection timeout.

src/main/java/com/feedzai/commons/sql/abstraction/engine/configuration/PdbProperties.java
```diff
@@ -150,6 +152,14 @@
      * space in the DB server. This depends on the database implementation.
      */
     public static final String DISABLE_LOB_CACHING = "pdb.disable_lob_caching";
+    /**
+     * Property that indicates the time of login timeout in the socket.
+     */
+    public static final String LOGIN_TIMEOUT = "pdb.login_timeout";
+    /**
+     * Property that indicates the time of login timeout in the socket.
```
nuno-azevedo: Copy pastey from above.

src/main/java/com/feedzai/commons/sql/abstraction/util/Constants.java
```diff
@@ -84,4 +84,12 @@
      * server. This is only possible on implementations that support this behavior. The default value is {@code false}.
      */
     public static final boolean DEFAULT_DISABLE_LOB_CACHING = false;
+    /**
+     * The default login socket timeout.
```
nuno-azevedo: login socket?

src/main/java/com/feedzai/commons/sql/abstraction/engine/configuration/PdbProperties.java
TiagoMSSantos: You are right. My bad. Thanks for mention it.

src/main/java/com/feedzai/commons/sql/abstraction/util/Constants.java
TiagoMSSantos: It's the timeout of making a connection to the socket. Because, the database can fail to connect. I will describe it better in javadocs.

lmtjalves: Suggestion: Default duration (in seconds) to wait for the database connection to be established.

TiagoMSSantos: Thanks for the suggestion. You are right. That way makes more sense.

src/main/java/com/feedzai/commons/sql/abstraction/engine/configuration/PdbProperties.java
```diff
@@ -150,6 +152,14 @@
      * space in the DB server. This depends on the database implementation.
      */
     public static final String DISABLE_LOB_CACHING = "pdb.disable_lob_caching";
+    /**
```
lmtjalves: new line before the javadoc

src/main/java/com/feedzai/commons/sql/abstraction/engine/configuration/PdbProperties.java
```diff
@@ -150,6 +152,14 @@
...
+     * Property that indicates the time, in seconds, of login timeout in the socket.
+     */
+    public static final String LOGIN_TIMEOUT = "pdb.login_timeout";
+    /**
```
lmtjalves: new line before the javadoc

src/main/java/com/feedzai/commons/sql/abstraction/engine/configuration/PdbProperties.java
```diff
@@ -150,6 +152,14 @@
      * space in the DB server. This depends on the database implementation.
      */
     public static final String DISABLE_LOB_CACHING = "pdb.disable_lob_caching";
+    /**
+     * Property that indicates the time, in seconds, of login timeout in the socket.
```
lmtjalves: I'm OK with keeping the property name as is, but I think there's missing some more context on the javadoc. For instance, something like: "... in seconds, to wait for the database connection to be established" is more easy to understand than "login timeout in the socket".

src/main/java/com/feedzai/commons/sql/abstraction/engine/AbstractDatabaseEngine.java
```diff
@@ -252,10 +261,11 @@ protected void connect() throws Exception {
 
         final String jdbc = getFinalJdbcConnection(this.properties.getJdbc());
 
-        this.conn = DriverManager.getConnection(
-                jdbc,
-                username,
-                password);
+        final Properties props = getDBProperties();
```
lmtjalves: Is this change needed?

src/main/java/com/feedzai/commons/sql/abstraction/engine/AbstractDatabaseEngine.java
```diff
@@ -221,6 +222,14 @@ protected String getPrivateKey() throws Exception {
         return readString(new FileInputStream(f)).trim();
     }
 
+    /**
```
lmtjalves: WDYT of having this instead:
```
Get the properties to be passed to the database connection that will be created. Since these properties are specific to the database engine, it is up to the {@link AbstractDatabaseEngine} implementation to specify them. This is used, for instance, to set the database connection timeout.
```

TiagoMSSantos: Yes, that's much better. Thanks for the feedback.

src/main/java/com/feedzai/commons/sql/abstraction/engine/configuration/PdbProperties.java
TiagoMSSantos: Already fixed.

src/main/java/com/feedzai/commons/sql/abstraction/engine/configuration/PdbProperties.java
TiagoMSSantos: Already fixed.

src/main/java/com/feedzai/commons/sql/abstraction/util/Constants.java
```diff
@@ -84,4 +84,14 @@
      * server. This is only possible on implementations that support this behavior. The default value is {@code false}.
      */
     public static final boolean DEFAULT_DISABLE_LOB_CACHING = false;
+
+    /**
+     * Default duration (in seconds) to wait for the database connection to be established.
+     */
+    public static final String DEFAULT_LOGIN_TIMEOUT = "0";
```
LuisEduardoReis: Explain in the javadoc why the default is 0 and what it means. I'm guessing it means there is no timeout for most databases.

TiagoMSSantos: Yes, you are right, 0 means no timeout. I will explain it there. Thanks for the feedback.

src/test/java/com/feedzai/commons/sql/abstraction/engine/impl/h2/H2EngineSchemaTest.java
```diff
@@ -128,4 +142,37 @@ public static int GetOne() {
...
+                setProperty(SCHEMA_POLICY, "create");
+                setProperty(SOCKET_TIMEOUT, Integer.toString(socketTimeout));
+                setProperty(LOGIN_TIMEOUT, Integer.toString(loginTimeout));
+            }
+        };
+
+        final DatabaseEngine de = DatabaseFactory.getConnection(properties);
+        final int connectionTimeoutInMs = de.getConnection().getNetworkTimeout();
+        // Not supported
+        assertEquals("timeout in db connection ok?", 0, connectionTimeoutInMs);
```
LuisEduardoReis: Error message could be more specific: "Is the timeout of the DB connection disabled as expected?"

src/test/java/com/feedzai/commons/sql/abstraction/engine/impl/abs/AbstractEngineSchemaTest.java
```diff
@@ -420,6 +425,38 @@ public void testInsertBatchRandomValuesDoNoWorkInBinaryDoubleColumn() throws Exc
...
+                setProperty(ENGINE, config.engine);
+                setProperty(SCHEMA_POLICY, "create");
+                setProperty(SOCKET_TIMEOUT, Integer.toString(socketTimeout));
+                setProperty(LOGIN_TIMEOUT, Integer.toString(loginTimeout));
+            }
+        };
+
+        final DatabaseEngine de = DatabaseFactory.getConnection(properties);
+        final int connectionTimeoutInMs = de.getConnection().getNetworkTimeout();
+        assertEquals("timeout in db connection ok?", socketTimeout * 1000, connectionTimeoutInMs);
```
LuisEduardoReis: Error message could be more specific: "Is the timeout of the DB connection the expected?"

TiagoMSSantos: Yes, that's better. Thanks.

src/test/java/com/feedzai/commons/sql/abstraction/engine/impl/h2/H2EngineSchemaTest.java
TiagoMSSantos: Yes, that is a special case where the socket timeout can't be set currently for the H2 and DB2 databases. Thanks for the feedback.

src/main/java/com/feedzai/commons/sql/abstraction/engine/configuration/PdbProperties.java
TiagoMSSantos: Already fixed. Thanks for the comment.

Issue #123, opened 2019-11-11
Add support for timeouts in read queries


leitaop:
In addition to the already existent network timeouts, add support for timeouts on query execution, i.e., expose Statement#setQueryTimeout() in the pdb API.

PR #124, opened 2019-11-13
Adds support for timeouts in select queries (#123)

leitaop:
This adds a configuration property (query_timeout) to specify a default timeout for all select queries created by a DatabaseEngine, allowing it to be overridden on individual queries. It also exposes the Statement#cancel() method, that allows a thread to cancel a query running on other thread.

This relies on the underlying JDBC driver supporting the Statement#setQueryTimeout()/cancel() methods, which is not supported by all database drivers.

### DIFF
--- a/src/main/java/com/feedzai/commons/sql/abstraction/engine/AbstractDatabaseEngine.java
+++ b/src/main/java/com/feedzai/commons/sql/abstraction/engine/AbstractDatabaseEngine.java
@@ -1124,7 +1124,7 @@ public abstract class AbstractDatabaseEngine implements DatabaseEngine {
      * @throws DatabaseEngineException If something goes wrong executing the query.
      */
     @Override
-    public synchronized List<Map<String, ResultColumn>> query(final Expression query) throws DatabaseEngineException {
+    public List<Map<String, ResultColumn>> query(final Expression query) throws DatabaseEngineException {
         return query(translate(query));
     }
 
@@ -1140,13 +1140,13 @@ public abstract class AbstractDatabaseEngine implements DatabaseEngine {
      * @throws DatabaseEngineException If something goes wrong executing the query.
      */
     @Override
-    public synchronized List<Map<String, ResultColumn>> query(final String query) throws DatabaseEngineException {
+    public List<Map<String, ResultColumn>> query(final String query) throws DatabaseEngineException {
         return processResultIterator(iterator(query));
     }
 
     @Override
     public List<Map<String, ResultColumn>> query(String query, int readTimeoutOverride) throws DatabaseEngineException {
-        return processResultIterator(iterator(query, DEFAULT_FETCH_SIZE, readTimeoutOverride));
+        return processResultIterator(iterator(query, properties.getFetchSize(), readTimeoutOverride));
     }
 
     /**
@@ -1156,7 +1156,7 @@ public abstract class AbstractDatabaseEngine implements DatabaseEngine {
      * @return A list of rows in the form of {@link Map}.
      * @throws DatabaseEngineException If a database access error occurs.
      */
-    protected List<Map<String, ResultColumn>> processResultIterator(ResultIterator it) throws DatabaseEngineException {
+    protected synchronized List<Map<String, ResultColumn>> processResultIterator(ResultIterator it) throws DatabaseEngineException {
         List<Map<String, ResultColumn>> res = new ArrayList<>();
 
         Map<String, ResultColumn> temp;
@@ -1168,12 +1168,12 @@ public abstract class AbstractDatabaseEngine implements DatabaseEngine {
     }
 
     @Override
-    public synchronized ResultIterator iterator(String query) throws DatabaseEngineException {
+    public ResultIterator iterator(String query) throws DatabaseEngineException {
         return iterator(query, properties.getFetchSize());
     }
 
     @Override
-    public synchronized ResultIterator iterator(Expression query) throws DatabaseEngineException {
+    public ResultIterator iterator(Expression query) throws DatabaseEngineException {
         return iterator(query, properties.getFetchSize());
     }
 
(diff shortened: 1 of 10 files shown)


src/test/java/com/feedzai/commons/sql/abstraction/engine/impl/abs/QueryTimeoutsTest.java
```diff
@@ -0,0 +1,189 @@
...
+    private static List<DatabaseEngineDriver> SUPPORTED_DRIVERS = ImmutableList.of(
+            DatabaseEngineDriver.POSTGRES,
+            DatabaseEngineDriver.MYSQL
+    );
+
+    /**
+     * A query that makes a sort in 1.000.000.000 rows, taking WAY more than
+     * the test timeout until the first row is returned.
+     */
+    private static Query HEAVY_QUERY = SqlBuilder
```
jmf-tls: suggestion: instead of this query, the some DBs have "sleep" functions that can be used in queries.
What do you think about something like
```
        switch (driver) {
            case POSTGRES:
                heavy = "SELECT pg_sleep(" + QUERY_MINIMUM_TIME + ")";
                break;
            case SQLSERVER:
                heavy = "WAITFOR DELAY '00:00:" + QUERY_MINIMUM_TIME + "'; SELECT 1";
                break;
            case MYSQL:
                heavy = "SELECT SLEEP(" + QUERY_MINIMUM_TIME + ")";
                break;
            default:
                assumeTrue(driver + " engine doesn't support setting timeouts, tests will be skipped", false);
```
where QUERY_MINIMUM_TIME would be the time to sleep in seconds

leitaop: Much better indeed!

src/main/java/com/feedzai/commons/sql/abstraction/engine/impl/MySqlEngine.java
```diff
@@ -861,14 +863,25 @@ protected void setSchema(final String schema) throws DatabaseEngineException {
     }
 
     @Override
-    public synchronized ResultIterator iterator(String query) throws DatabaseEngineException {
+    public ResultIterator iterator(String query) throws DatabaseEngineException {
+        return iterator(query, properties.getFetchSize(), properties.getSelectQueryTimeout());
+    }
```
jmf-tls: I suppose you kept this one because the super method has a default fetch size of DEFAULT_FETCH_SIZE instead of `properties.getFetchSize()`?
Turns out that didn't seem very correct, and was fixed.
If you rebase to the head of "master" branch, I think calling the `super.iterator(String query)` will end up using the same fetch size and select query timeout as this method, so the override won't be needed anymore.

src/main/java/com/feedzai/commons/sql/abstraction/engine/impl/MySqlEngine.java
```diff
@@ -861,14 +863,25 @@ protected void setSchema(final String schema) throws DatabaseEngineException {
...
+    @Override
+    public ResultIterator iterator(String query, int fetchSize, int readTimeout) throws DatabaseEngineException {
         try {
             getConnection();
             final Statement stmt = conn.createStatement(TYPE_FORWARD_ONLY, CONCUR_READ_ONLY);
-            stmt.setFetchSize(properties.getFetchSize());
+            stmt.setFetchSize(fetchSize);
+            if (readTimeout != NO_TIMEOUT) {
+                stmt.setQueryTimeout(readTimeout);
+            }
```
jmf-tls: I think that with the changes in #127 (replace DEFAULT_FETCH_SIZE in iterator methods with value from config) and with your changes, this method can be exactly the same as the super method if instead you override `createSelectStatement` to use `conn.createStatement(TYPE_FORWARD_ONLY, CONCUR_READ_ONLY);`.

However, in the MySQL driver those values are the already default (they were default also when this change was introduced), so I think we can simply use `conn.createStatement()`, which makes both `iterator` methods overrides unnecessary at all in MySqlEngine!

Note: this override was introduced in a commit to "Fixes MySqlEngine not being able to iterate using fetch size.", but from my experiments because of CockroachDB, MySQL still doesn't respect the fetch size setting; it seems the only way to make it work is to use server-side cursors, enabled with a flag in the JDBC URL.

src/main/java/com/feedzai/commons/sql/abstraction/engine/AbstractDatabaseEngine.java
```diff
@@ -1154,23 +1182,36 @@ public synchronized ResultIterator iterator(Expression query) throws DatabaseEng
...
+        return iterator(translate(query), fetchSize);
+    }
+
+    @Override
+    public ResultIterator iterator(Expression query, int fetchSize, int readTimeoutOverride) throws DatabaseEngineException {
+        return iterator(translate(query), fetchSize, readTimeoutOverride);
+    }
+
+    @Override
+    public ResultIterator iterator(String query, int fetchSize, int readTimeoutOverride) throws DatabaseEngineException {
```
jmf-tls: The first `iterator` method in AbstractDatabaseEngine had this signature:  
```
public synchronized ResultIterator iterator(String query)
```
Then people added more iterator methods, some of them `synchronized`, some of them not.
I think all methods that use the connection should be `synchronized`; this means this one would need to be `synchronized`, but all other `iterator` methods call this one and don't need to be.

leitaop:
> Did I miss something?
> Anyway, it seems appropriate to check whether an Exception is an SQLTimeoutException by default, instead of always returning false, like I mentioned in a comment.

I changed the default to check for an SQLTimeoutException, are you looking at the right version ?

src/main/java/com/feedzai/commons/sql/abstraction/engine/AbstractDatabaseEngine.java
leitaop: Yep, removed some unnecessary synchronized. Ensured the processResultIterator was kept synchronized because it iterates in the result set, and that might involve fetching rows from the server.

Issue #126, opened 2019-11-18
Improvements from CockroachDB testing


jmf-tls:
The initial testing with CockroachDB resulted in finding some possible improvements, which apply to one or many of the existing DB engines:

1. allow read-only connections: allow a PDB configuration flag to specify that ALL queries in a given DB engine are going to be read-only; this information can be used to enable some optimizations in the DB drivers by setting `java.sql.Connection.setReadOnly(true)` - further optimizations can be done based on this, for e.g. setting transactions as read only, "as of timestamp" and/or read from replica, especially important in CockroachDB

1. **(done in #127)** AbstractDatabaseEngine has a constant DEFAULT_FETCH_SIZE, but that already exists with the same value in PdbProperties; besides that, the `iterator` methods in AbstractDatabaseEngine where a fetch size isn't provided, shouldn't be using a hardcoded value, but instead should use the configured value, whether it is the default or not;

1. **(done in #131)** change "allowColumnDrop" default to `false`: having the default at `true`, although convenient to keep the DB tables synchronized with the respective PDB entities, may lead to unintended loss of data;

1. fix fetchSize usage: PostgreSQL only respects this setting inside transactions; MySQL only respects fetchSize if "useCursorFetch" property is enabled (e.g. in connection properties or JDBC URL; default is false);

1. change addEntity to add primary keys, indexes, etc along with the CREATE statement, instead of creating a table and then altering it to add those indexes (besides the benefit of executing the creating DDL in a single step, this is a requirement for CockroachDB, which doesn't allow creating tables without a primary key and also doesn't allow changing it after creation);

1. add autoInc option to allow selecting the type to be used (maybe rename the parameter to "sequence"?); we may care that a certain column is monotonically increasing (and the current code may not always guarantee this), but other times we may just need a unique identifier for the row, which can be random and is lighter for the DB to produce;

1. **(done in #128)** unlike all other DB engines, deadlocks may occur while using SQL Server and DB locks are interleaved with Java locks; this is because even in READ COMMITED, a SELECT on a row that is being written by another transaction will wait for it to be committed or rolled back;
[...]

PR #127, opened 2019-11-18
Use configured fetch size by default in iterators (originated by #126)

jmf-tls:
Summary:
Iterators were using a default fetch size defined in a constant in
 AbstractDatabaseEngine.
PDB configuration already has a fetch size property, which by default
 has the same value as that constant - this value is configurable and
 should be used instead of the constant.

### DIFF
--- a/src/main/java/com/feedzai/commons/sql/abstraction/engine/AbstractDatabaseEngine.java
+++ b/src/main/java/com/feedzai/commons/sql/abstraction/engine/AbstractDatabaseEngine.java
@@ -97,10 +97,7 @@ public abstract class AbstractDatabaseEngine implements DatabaseEngine {
      */
     @Inject
     protected AbstractTranslator translator;
-    /**
-     * The default fetch size for iterators.
-     */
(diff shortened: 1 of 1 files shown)


nmldiegues:
In the past I have had situations where changing the fetchSize seemed to have taken effect, but according to this review it shouldn't as the configuration was not being used?  Or are those methods overridden in any way? Or the configurable fetchSize is also being used in more places maybe?

Anyway, great simple improvement!

jmf-tls:
> In the past I have had situations where changing the fetchSize seemed to have taken effect, but according to this review it shouldn't as the configuration was not being used? Or are those methods overridden in any way? Or the configurable fetchSize is also being used in more places maybe?
> 
Yes, there are usages a the configured fetch size:
- the one from PDB properties is used in MySQL engine (where it currently doesn't work anyway -- see point 4 of #126 :smiling_imp: )
- in public some public _iterator()_ methods where fetch size is a parameter (which I just noticed that, unlike the majority of public methods that talk to the DB, is not synchronized -- this may be a problem)

PR #128, opened 2019-11-20
Support custom isolation levels, add tests (originated by #126)

jmf-tls:
Summary:
SQL Server has a default behavior for certain isolation levels different
 from other DB engines: with READ COMMITTED or stricter, it will block
 on reads when another transaction is changing (but not committed yet)
 the rows being read; other engines simply return the currently
 committed value.

To make SQL Server have the same behavior, the current database must
 have enabled the flag READ_COMMITTED_SNAPSHOT.
With SERIALIZABLE isolation level the behavior isn't changed, the
 TRANSACTION SNAPSHOT level must be used instead (which is equivalent to
 SERIALIZABLE in PostgreSQL and Oracle) - this requires the current
 database to have the flag ALLOW_SNAPSHOT_ISOLATION enabled.

To allow and test these behaviors, the following changes were made in
 this commit:
- PdbProperties now allows setting custom numeric isolation levels (for
 example, setting it to "4096" will set TRANSACTION SNAPSHOT level for
 SQL Server;
- the SQL Server container that is used for tests now has a new database
 with the flags READ_COMMITTED_SNAPSHOT and ALLOW_SNAPSHOT_ISOLATION
 enabled (this was also reflected in the test connections.properties);
- new tests were created, to test that the custom TRANSACTION SNAPSHOT
 isolation level works, and check that a deadlock doesn't occur in a
 specific scenario.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -549,7 +549,7 @@
                         <configuration>
                             <images>
                                 <image>
-                                    <name>microsoft/mssql-server-linux:2017-CU13</name>
+                                    <name>mcr.microsoft.com/mssql/server:2017-CU17</name>
                                     <alias>pdb-sqlserver</alias>
                                     <run>
                                         <namingStrategy>alias</namingStrategy>
@@ -564,6 +564,16 @@
                                         <wait>
                                             <time>40000</time>
                                             <log>Service Broker manager has started</log>
+                                            <exec>
+                                                <!--
+                                                Prepare a new "pdb" database for tests with READ_COMMITTED_SNAPSHOT enabled
+                                                 (required for SQL server to have the same behavior as other DBs in READ_COMMITTED
+                                                 isolation level) and ALLOW_SNAPSHOT_ISOLATION enabled to allow using the
+                                                 TRANSACTION_SNAPSHOT level (equivalent to SERIALIZABLE in PostgreSQL and Oracle).
+                                                A new database needs to be used because the flags can't be enabled on "master" database.
+                                                -->
+                                                <postStart>/opt/mssql-tools/bin/sqlcmd -l 45 -U sa -P AaBb12.# -Q CREATE\ DATABASE\ pdb;\ ALTER\ DATABASE\ pdb\ SET\ READ_COMMITTED_SNAPSHOT\ ON;\ ALTER\ DATABASE\ pdb\ SET\ ALLOW_SNAPSHOT_ISOLATION\ ON</postStart>
+                                            </exec>
                                         </wait>
                                     </run>
                                 </image>
--- a/src/test/java/com/feedzai/commons/sql/abstraction/engine/impl/abs/EngineIsolationTest.java
+++ b/src/test/java/com/feedzai/commons/sql/abstraction/engine/impl/abs/EngineIsolationTest.java
@@ -15,8 +15,12 @@
  */
 package com.feedzai.commons.sql.abstraction.engine.impl.abs;
 
+import com.feedzai.commons.sql.abstraction.ddl.DbEntity;
+import com.feedzai.commons.sql.abstraction.engine.DatabaseEngine;
+import com.feedzai.commons.sql.abstraction.engine.DatabaseEngineDriver;
 import com.feedzai.commons.sql.abstraction.engine.DatabaseFactory;
 import com.feedzai.commons.sql.abstraction.engine.DatabaseFactoryException;
+import com.feedzai.commons.sql.abstraction.engine.configuration.PdbProperties;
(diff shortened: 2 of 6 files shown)


pom.xml
```diff
@@ -564,6 +564,16 @@
...
                                             <log>Service Broker manager has started</log>
+                                            <exec>
+                                                <!--
+                                                Prepare a new "pdb" database for tests with READ_COMMITTED_SNAPSHOT enabled
+                                                 (required for SQL server to have the same behavior as other DBs in READ_COMMITTED
+                                                 isolation level) and ALLOW_SNAPSHOT_ISOLATION enabled to allow using the
+                                                 TRANSACTION_SNAPSHOT level (equivalent to SERIALIZABLE in PostgreSQL and Oracle).
+                                                A new database needs to be used because the flags can't be enabled on "master" database.
+                                                -->
+                                                <postStart>/opt/mssql-tools/bin/sqlcmd -l 45 -U sa -P AaBb12.# -Q CREATE\ DATABASE\ pdb;\ ALTER\ DATABASE\ pdb\ SET\ READ_COMMITTED_SNAPSHOT\ ON;\ ALTER\ DATABASE\ pdb\ SET\ ALLOW_SNAPSHOT_ISOLATION\ ON</postStart>
```
nmldiegues: What's the advantage of doing this here vs in some initializer of the tests in Java?
Here feels more like magic that developers will tend to forget about easily.

pom.xml
```diff
@@ -564,6 +564,16 @@
                                         <wait>
                                             <time>40000</time>
                                             <log>Service Broker manager has started</log>
```
nmldiegues: service broker? maybe this was copied by accident

src/test/java/com/feedzai/commons/sql/abstraction/engine/impl/abs/EngineIsolationTest.java
```diff
@@ -77,13 +101,189 @@ public void readUncommittedTest() throws DatabaseFactoryException {
...
     public void repeatableReadTest() throws DatabaseFactoryException {
-        this.properties.setProperty(ISOLATION_LEVEL, "read_uncommitted");
+        this.properties.setProperty(ISOLATION_LEVEL, "repeatable_read");
         DatabaseFactory.getConnection(properties);
     }
 
     @Test
     public void serializableTest() throws DatabaseFactoryException {
-        this.properties.setProperty(ISOLATION_LEVEL, "read_uncommitted");
+        this.properties.setProperty(ISOLATION_LEVEL, "serializable");
```
nmldiegues: these were just plain wrong? they are unrelated to the change, right?

jmf-tls:
> Damn these DB guys who cannot comply with standards. Only if there was a SQL standard... oh there is, they just ignore it :)

Well, JDBC has a java.sql.Types.BOOLEAN, but Oracle's own database doesn't support booleans: the column needs to be of numeric or string type, to store 0s or 1s :smile:

src/test/java/com/feedzai/commons/sql/abstraction/engine/impl/abs/EngineIsolationTest.java
jmf-tls: Yes, I noticed this because of the new SQL server "transaction snapshot" isolation level.
They were wrong in the sense that the isolation level simply didn't match what the test name implies - the tests were working but testing the wrong values :smile: (since ever)

pom.xml
jmf-tls: It's intended, this is one of the last lines that appears in the log when SQL server starts:
`2019-11-22 11:40:46.02 spid21s     Service Broker manager has started.`
We assume that SQL server is started when we get that...
They call it "broker"? :man_shrugging:

pom.xml
jmf-tls: Let's call it "transparent" to the user/developer.
The docker initialization performed here reflects how a production database and PDB should work together -- in this case of SQL Server, PDB is expected to work with a non-"master" database (which is recommended anyway) that needs to have at least READ_COMMITTED_SNAPSHOT enabled in order to work like the other DB engines.
This same requirement is mentioned in the "quirks" page in the PDB wiki.

This is transparent in the sense that, while these settings in the pom are not readily accessible from Java code and are easy to forget about, in reality they only need to be in sync with the contents of the "connections.properties" file (usernames, passwords, database, etc.) to be used in the tests.
In this particular case, creating the "pdb" database would require some code to run before all the tests, establishing an initial connection to the "master" database just to create the new one (note that we're using the vanilla docker containers and we're not using doker-java :smile: ).

PR #129, opened 2019-11-28
Makes query timeout support non-breaking

leitaop:
The query timeout feature added some methods to the DatabaseEngine interface that introduce
an API change on classes that implement it (typically wrapper classes). These methods might
not be relevant in all contexts and so a default implementation where the read timeout is
ignored was added.

### DIFF
--- a/src/main/java/com/feedzai/commons/sql/abstraction/engine/DatabaseEngine.java
+++ b/src/main/java/com/feedzai/commons/sql/abstraction/engine/DatabaseEngine.java
@@ -304,13 +304,16 @@ public interface DatabaseEngine extends AutoCloseable {
     List<Map<String, ResultColumn>> query(final Expression query) throws DatabaseEngineException;
 
     /**
-     * Executes the given query overriding the configured query timeout (see {@link PdbProperties#getSelectQueryTimeout()}).
+     * Executes the given query overriding the configured query timeout (see {@link PdbProperties#getSelectQueryTimeout()})
+     * if the underlying engine supports query timeouts.
      *
      * @param query The query to execute.
      * @param readTimeoutOverride The query timeout to use.
      * @throws DatabaseEngineException If something goes wrong executing the query.
      */
-    List<Map<String, ResultColumn>> query(final Expression query, final int readTimeoutOverride) throws DatabaseEngineException;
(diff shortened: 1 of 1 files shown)


src/main/java/com/feedzai/commons/sql/abstraction/engine/DatabaseEngine.java
```diff
@@ -614,15 +623,18 @@ void createPreparedStatement(final String name, final String query, final int ti
...
      *
      * @param query     The expression that represents the query.
      * @param fetchSize The number of rows to fetch each time.
      * @param readTimeoutOverride The query timeout to use.
      * @return An iterator for the results of the given SQL expression.
      * @throws DatabaseEngineException If a database access error occurs.
      */
-    ResultIterator iterator(final Expression query, final int fetchSize, final int readTimeoutOverride) throws DatabaseEngineException;
+    default ResultIterator iterator(final Expression query, final int fetchSize, final int readTimeoutOverride) throws DatabaseEngineException {
+        return iterator(query, fetchSize);
```
jmf-tls: since `translate` is available as an interface method, you could instead use what the abstract engine is using:
```
        return iterator(translate(query), fetchSize, readTimeoutOverride);
```

src/main/java/com/feedzai/commons/sql/abstraction/engine/DatabaseEngine.java
```diff
@@ -304,13 +304,16 @@
...
+     * Executes the given query overriding the configured query timeout (see {@link PdbProperties#getSelectQueryTimeout()})
+     * if the underlying engine supports query timeouts.
      *
      * @param query The query to execute.
      * @param readTimeoutOverride The query timeout to use.
      * @throws DatabaseEngineException If something goes wrong executing the query.
      */
-    List<Map<String, ResultColumn>> query(final Expression query, final int readTimeoutOverride) throws DatabaseEngineException;
+    default List<Map<String, ResultColumn>> query(final Expression query, final int readTimeoutOverride) throws DatabaseEngineException {
+        return query(query);
```
jmf-tls: since `translate` is available as an interface method, you could instead use what the abstract engine is using:
```
        return query(translate(query), readTimeoutOverride);
```

PR #132, opened 2019-11-29
Add support for CockroachDB

jmf-tls:
Summary:
CockroachDB uses the same driver as PostgreSQL, but there are some
incompatibilities, and a new profile needed to be setup for tests to
allow using a CockroachDB Docker image.

This commit adds a new CockroachDB engine that extends the one for
PostgreSQL to address the incompatibilities and makes some small
adjustments to all code accordingly.
A new "cockroach" profile was added to the pom to run the tests with a
CockroachDB instance.

The pom was changed by adding new "image.<profile_name>" flags, one for
each profile, that allows specifying docker images different from the
defaults (e.g. new versions) to run the tests with.
Those defaults are the images/versions that were already in the pom.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -486,6 +486,9 @@
 
         <profile>
             <id>mysql</id>
+            <properties>
+                <image.mysql>mysql:5.7.22</image.mysql>
+            </properties>
             <build>
                 <plugins>
                     <plugin>
@@ -504,7 +507,7 @@
                         <configuration>
                             <images>
                                 <image>
-                                    <name>mysql:5.7.22</name>
+                                    <name>${image.mysql}</name>
                                     <alias>pdb-mysql</alias>
                                     <run>
                                         <namingStrategy>alias</namingStrategy>
@@ -531,6 +534,9 @@
 
         <profile>
             <id>sqlserver</id>
+            <properties>
+                <image.sqlserver>mcr.microsoft.com/mssql/server:2017-CU17</image.sqlserver>
+            </properties>
             <build>
                 <plugins>
                     <plugin>
@@ -549,7 +555,7 @@
                         <configuration>
                             <images>
                                 <image>
-                                    <name>mcr.microsoft.com/mssql/server:2017-CU17</name>
+                                    <name>${image.sqlserver}</name>
                                     <alias>pdb-sqlserver</alias>
                                     <run>
                                         <namingStrategy>alias</namingStrategy>
@@ -586,6 +592,9 @@
 
         <profile>
             <id>postgresql</id>
+            <properties>
+                <image.postgresql>postgres:9.6.14</image.postgresql>
+            </properties>
             <build>
                 <plugins>
                     <plugin>
@@ -604,7 +613,7 @@
                         <configuration>
                             <images>
                                 <image>
-                                    <name>postgres:9.6.14</name>
+                                    <name>${image.postgresql}</name>
                                     <alias>pdb-postgresql</alias>
                                     <run>
                                         <namingStrategy>alias</namingStrategy>
@@ -629,7 +638,10 @@
         </profile>
 
         <profile>
-            <id>postgresql11</id>
+            <id>cockroach</id>
+            <properties>
+                <image.cockroach>cockroachdb/cockroach:v19.2.0</image.cockroach>
+            </properties>
             <build>
                 <plugins>
                     <plugin>
@@ -637,8 +649,9 @@
                         <artifactId>maven-surefire-plugin</artifactId>
                         <configuration>
                             <systemProperties>
-                                <instances>postgresql</instances>
+                                <instances>cockroach</instances>
                             </systemProperties>
+                            <excludedGroups>com.feedzai.commons.sql.abstraction.engine.impl.cockroach.SkipTestCockroachDB</excludedGroups>
                         </configuration>
                     </plugin>
 
@@ -648,20 +661,19 @@
                         <configuration>
                             <images>
                                 <image>
-                                    <name>postgres:11.4</name>
-                                    <alias>pdb-postgresql11</alias>
+                                    <name>${image.cockroach}</name>
+                                    <alias>pdb-cockroach</alias>
                                     <run>
                                         <namingStrategy>alias</namingStrategy>
-                                        <hostname>pdb-postgresql11</hostname>
+                                        <hostname>pdb-cockroach</hostname>
                                         <ports>
-                                            <port>5432:5432</port>
+                                            <port>8080:8080</port>
+                                            <port>26257:26257</port>
                                         </ports>
-                                        <env>
-                                            <POSTGRES_PASSWORD>AaBb12.#</POSTGRES_PASSWORD>
-                                        </env>
+                                        <cmd>start --insecure</cmd>
                                         <wait>
                                             <time>40000</time>
-                                            <log>PostgreSQL init process complete; ready for start up</log>
+                                            <log>CockroachDB node starting</log>
                                         </wait>
                                     </run>
                                 </image>
@@ -674,6 +686,9 @@
 
         <profile>
             <id>oracle</id>
+            <properties>
+                <image.oracle>iatebes/oracle_11g:latest</image.oracle>
+            </properties>
             <build>
                 <plugins>
                     <plugin>
@@ -697,7 +712,7 @@
                         <configuration>
                             <images>
                                 <image>
-                                    <name>iatebes/oracle_11g:latest</name>
+                                    <name>${image.oracle}</name>
                                     <alias>pdb-oracle</alias>
                                     <run>
                                         <namingStrategy>alias</namingStrategy>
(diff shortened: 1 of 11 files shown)


src/main/java/com/feedzai/commons/sql/abstraction/engine/AbstractDatabaseEngine.java
```diff
@@ -701,8 +701,8 @@ public synchronized void dropEntity(String entity) throws DatabaseEngineExceptio
      */
     @Override
     public synchronized void dropEntity(final DbEntity entity) throws DatabaseEngineException {
-        dropSequences(entity);
```
leitaop: Is this change in the order intentional ?  If yes, please leave a comment indicating why sequences have to be dropped after tables.

src/main/java/com/feedzai/commons/sql/abstraction/engine/impl/CockroachDBEngine.java
```diff
@@ -0,0 +1,384 @@
...
+
+            long ret = 0;
+
+            if (useAutoInc) {
+                if (generatedKeys.next()) {
+                    ret = generatedKeys.getLong(1);
+                }
+            } else if (hasIdentityColumn(me.getEntity())) {
+                final List<Map<String, ResultColumn>> q = query(
+                        "SELECT max(" + quotize(me.getAutoIncColumn()) + ") FROM" + quotize(name)
```
joaonuno: Isn't this missing a space after FROM? Is this being tested? Using format seems a better choice here.

src/main/java/com/feedzai/commons/sql/abstraction/engine/impl/PostgreSqlEngine.java
```diff
@@ -113,6 +113,21 @@ public PostgreSqlEngine(PdbProperties properties) throws DatabaseEngineException
...
+     *
+     * Note: this constructor exists so that most of the code in this implementation can be reused
+     * by (almost) compatible engines, like CockroachDB.
+     *
+     * @param properties The properties for the database connection.
+     * @param driver     The driver to connect to the database.
+     * @throws DatabaseEngineException When the connection fails.
+     * @since 2.5.0
+     */
+    public PostgreSqlEngine(final PdbProperties properties, final String driver) throws DatabaseEngineException {
```
leitaop: protected ?

src/main/java/com/feedzai/commons/sql/abstraction/engine/impl/CockroachDBTranslator.java
```diff
@@ -0,0 +1,137 @@
...
+    public String translate(final DbColumn c) {
+        switch (c.getDbColumnType()) {
+            case BOOLEAN:
+                return "BOOLEAN";
+
+            case DOUBLE:
+                return "DOUBLE PRECISION";
+
+            case INT:
+                if (c.isAutoInc()) {
```
leitaop: Condition seems unnecessary since INT4 is returned any way.

src/main/java/com/feedzai/commons/sql/abstraction/engine/impl/CockroachDBTranslator.java
```diff
@@ -0,0 +1,137 @@
...
+                    return "INT4";
+                } else {
+                    return "INT4";
+                }
+
+            case LONG:
```
leitaop: Same as above

src/main/java/com/feedzai/commons/sql/abstraction/engine/impl/CockroachDBEngine.java
```diff
@@ -0,0 +1,384 @@
...
+     *
+     * @param properties The properties for the database connection.
+     * @throws DatabaseEngineException When the connection fails.
+     */
+    public CockroachDBEngine(final PdbProperties properties) throws DatabaseEngineException {
+        super(properties, COCKROACHDB_DRIVER);
+    }
+
+    @Override
+    protected void createTable(final DbEntity entity) throws DatabaseEngineException {
```
leitaop: There's a lot of code here that is duplicate of the one in the Posgres engine and that could be eliminated with a minor refactor. Doing that would help to understand what is different between Posgres and Cockroach.

pom.xml
```diff
@@ -629,16 +638,20 @@
         </profile>
 
         <profile>
-            <id>postgresql11</id>
```
leitaop: Seems we're replacing the PG 11 profile by the cockroach one. Wouldn't it make more sense to replace the 9.6 (somewhat obsolete) one instead ?

marpereira: Here we are removing specific profiles for each database engine. Giving flexibility to define engine version (docker image) throw maven parameter.

leitaop: My point was that <image.postgresql>postgres:9.6.14</image.postgresql> points to an obsolete driver, the default for the "postgres" profile should be 11.

marpereira: Didn't understand the main issue that you point. Your right, and in that case, de default image should be latest, instead of a "today latest x.y.z version".

src/main/java/com/feedzai/commons/sql/abstraction/engine/AbstractDatabaseEngine.java
jmf-tls: It's intentional, I'll add a comment.
Most of the engines don't even use sequences and `dropSequences` is a no-op for them.
DB2 and Oracle use them, but only as the value for the autoinc column in INSERT prepared statements.

CockroachDB uses autoinc like PostgreSQL, but the default is to translate SERIAL as a unique  numeric ID, to have the autoinc functionality a sequence had to be created; however, in this case, there is a dependency on that sequence from the table that uses it, preventing the sequence from being dropped.

src/main/java/com/feedzai/commons/sql/abstraction/engine/impl/CockroachDBEngine.java
jmf-tls: Nice catch!
For now, this is only a cosmetic issue, it works because what follows is always quotized and the quotes also act as a separator (but the space should be there anyway).

src/main/java/com/feedzai/commons/sql/abstraction/engine/impl/CockroachDBTranslator.java
jmf-tls: Good point :smile: 
Maybe later we will want to make a distinction here if we just want a generated unique number, but if we want an autoinc in CockroachDB similar to the other engines, then the translation here is the same, autoinc or not.

src/main/java/com/feedzai/commons/sql/abstraction/engine/impl/CockroachDBEngine.java
jmf-tls: You're correct - and it is planned (issue #126).
CockroachDB needs to have a primary key created along with the table, otherwise a default one is created, and afterwards it is not possible to add another one or replace the default.
The other engines don't need it, but they can also have a CREATE TABLE statement that also defines primary key, indexes, etc; when this is changed for PostgreSQL, then the override for CockroachDB probably can be removed.

pom.xml
jmf-tls: End of life for PostgreSQL is in 2021, and we support it in our products.
But the reason the postgresql11 profile was removed is that now it is possible to choose the Docker image for tests.
For Postgres 11, the flags could be `-Ppostgresql -Dimage.postgresql=postgres:11.4`, and similarly we could choose an image for Postgres 10 or even 12.

jmf-tls: PS: just noticed the comments from Mario :smile: 
To clarify, the reason for removing postgresql11 is that we can use it, or even a newer version, using the image version flag; the reason for keeping 9.6 as the default is that both PostgreSQL and us in our products still support that version (and we don't want to introduce support for new features from v11.x that won't work in the older v9.x)

src/main/java/com/feedzai/commons/sql/abstraction/engine/impl/CockroachDBEngine.java
leitaop: Please ensure that is documented in issue #126 then, didn't see any reference to this.

pom.xml
leitaop: I don't agree that the version for the "postgres" profile is PG 9 (the one that people use without explicitly overriding the image) but I'm not blocking this review for this.

src/main/java/com/feedzai/commons/sql/abstraction/engine/impl/CockroachDBEngine.java
jmf-tls: It's on point 5.
#126 is a bit messy, each of those things should have it's own ticket#, but from the introduction of CockroachDB these issues kept coming up and it was easier to just add more to an existing list than creating a new issue for each one :smile_cat:

pom.xml
leitaop: Oops, sorry, was confusing with Oracle ...

PR #133, opened 2019-12-04
Fix test broken after "allowColumnDrop" default change

jmf-tls:
Summary:
On #131 the default for "allowColumnDrop" was changed to false.
This broke the test "testDefaultAllowColumnDrop", this commit
fixes the test according to the new default value and splits it
into 2 tests, one just to test allowColumnDrop default and the
other to test when the value is set to true.


jmf-tls:
This was reviewed by Mário Pereira, Bruno Flávio and Miguel Pereira.

PR #134, opened 2019-12-13
Cleanups in preparation for more retryable exceptions

jmf-tls:
Summary:
Retryable errors when running queries may occur in other places
besides the commit.

In preparation for that, the method
 `persist(final String name, final EntityEntry entry)`
was refactored to AbstractDatabaseEngine (it was implemented in
all engines, but the code was always the same).

This way, any try-catch for the retryable exceptions only needs
to be done in the persist with autoInc parameter.

While at it, the method `translateType(DbColumn c)` was also
repeated in all engines and was moved to AbstractDatabaseEngine.

Other cleanups highlighted by the IDE were also done.

### DIFF
--- a/src/main/java/com/feedzai/commons/sql/abstraction/engine/AbstractDatabaseEngine.java
+++ b/src/main/java/com/feedzai/commons/sql/abstraction/engine/AbstractDatabaseEngine.java
@@ -1228,7 +1228,7 @@ public abstract class AbstractDatabaseEngine implements DatabaseEngine {
     protected abstract ResultIterator createResultIterator(Statement statement, String sql) throws DatabaseEngineException;
 
     /**
-     * Creates a specific {@link ResultIterator} for the engine in place given given prepared statement.
+     * Creates a specific {@link ResultIterator} for the engine in place given a prepared statement.
      *
      * @param ps The prepared statement.
      * @return The result iterator.
--- a/src/main/java/com/feedzai/commons/sql/abstraction/engine/AbstractDatabaseEngine.java
+++ b/src/main/java/com/feedzai/commons/sql/abstraction/engine/AbstractDatabaseEngine.java
@@ -471,13 +471,15 @@ public abstract class AbstractDatabaseEngine implements DatabaseEngine {
             final PreparedStatement insertReturning = mappedEntity.getInsertReturning();
             final PreparedStatement insertWithAutoInc = mappedEntity.getInsertWithAutoInc();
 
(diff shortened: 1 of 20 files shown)


src/main/java/com/feedzai/commons/sql/abstraction/engine/AbstractDatabaseEngine.java
```diff
@@ -471,13 +471,15 @@ private void closeMappedEntity(final MappedEntity mappedEntity) {
             final PreparedStatement insertReturning = mappedEntity.getInsertReturning();
             final PreparedStatement insertWithAutoInc = mappedEntity.getInsertWithAutoInc();
 
-            insert.executeBatch();
+            if (!insert.isClosed()) {
```
TiagoMSSantos: Don't we need to do like the others?
`if (insert != null && !insert.isClosed())`

src/main/java/com/feedzai/commons/sql/abstraction/engine/AbstractDatabaseEngine.java
```diff
@@ -617,7 +619,7 @@ public synchronized void updateEntity(DbEntity entity) throws DatabaseEngineExce
                 }
                 if (!toRemove.isEmpty()) {
                     if (properties.allowColumnDrop()) {
-                        dropColumn(entity, toRemove.toArray(new String[toRemove.size()]));
+                        dropColumn(entity, toRemove.toArray(new String[0]));
```
TiagoMSSantos: Doing `new String[0]`  means that `toArray` method will allocate an array with the right size, right?

src/main/java/com/feedzai/commons/sql/abstraction/engine/AbstractDatabaseEngine.java
```diff
@@ -1224,6 +1227,15 @@ public synchronized ResultIterator iterator(String query, int fetchSize, int rea
      */
     protected abstract ResultIterator createResultIterator(Statement statement, String sql) throws DatabaseEngineException;
 
+    /**
+     * Creates a specific {@link ResultIterator} for the engine in place given given prepared statement.
```
joaonuno: given * 2

jmf-tls: :smile: 
This method (incl. javadoc) was not modified, just moved from below to be close to the other `createResultIterator`

but nice catch!

PR #136, opened 2019-12-18
Throw retryable exceptions in the methods where they can occur

jmf-tls:
Summary:
A previous commit added a flag/method on DatabaseEngineRuntimeException
to indicate that the error that occurred was retryable.

However, deadlocks may occur not only on commits but also on any other
operation in a transaction, even in reads. In some cases a deadlocks
may even occur on queries not running in transactions.

This commit adds SQL exception handling to the other methods, so that
if the action is retryable, a corresponding "retryable" exception will
be thrown to indicate that.

Tests were added to verify that when deadlocks occur an appropriate
"retryable" exception is thrown and a rollback+retry solves the
problem.


jmf-tls:
continued from #135

PR #139, opened 2020-05-12
Fix computed column boolean conversion

joaonuno:
In MySql and SqlServer, the result column  failed
when called on a computed column of boolean type. This is caused
by the implicit type conversions that the driver does when fetching
a value. For regular columns, the type is well known and the
conversion between 0/1 to false/true works. For computed columns,
the type conversions fail because the driver doesn't know the type.

See the new unit-test a simple scenario that failed. Also see the
new test Javadoc for more details on another (failed) approach.

### DIFF
--- a/src/main/java/com/feedzai/commons/sql/abstraction/dml/result/SqlServerResultColumn.java
+++ b/src/main/java/com/feedzai/commons/sql/abstraction/dml/result/SqlServerResultColumn.java
@@ -15,8 +15,10 @@
  */
 package com.feedzai.commons.sql.abstraction.dml.result;
 
+import com.feedzai.commons.sql.abstraction.engine.DatabaseEngineRuntimeException;
+
 /**
- * The PostgreSql column result implementation.
+ * The Sql Server column result implementation.
  *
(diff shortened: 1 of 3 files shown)


src/main/java/com/feedzai/commons/sql/abstraction/dml/result/SqlServerResultColumn.java
```diff
@@ -32,4 +34,27 @@
...
+
+        if (val.equals("1")) {
+            return true;
+        }
+
+        if (val.equals("0")) {
+            return false;
+        }
+
+        throw new DatabaseEngineRuntimeException(val + " is not a boolean type");
```
francisco-polaco, line 58: nitpick: I would prefer a if-then-else approach, but I'm nitpicking so it's just a suggestion. :)

joaonuno: I'll leave it this way to keep consistency with the other `ResultColumn` implementations that already follow this pattern.

PR #140, opened 2020-06-03
Add support to drop a view.

nuno-santos96:
Added dropView method, similar to the existing createView.


nuno-santos96:
Tests ran on all DBs. All green.

src/main/java/com/feedzai/commons/sql/abstraction/dml/dialect/SqlBuilder.java
```diff
@@ -560,6 +560,16 @@ public static View createView(final String name) {
         return new View(name);
     }
 
+    /**
+     * Drops a view.
+     *
+     * @param name The name of the view.
+     * @return The drop view representation.
+     */
+    public static DropView dropView(final String name) {
```
joaonuno, line 569: The other drop* commands are defined in `AbstractDatabaseEngine`. Did you consider adding this one there also? What would be the pros/cons?

src/main/java/com/feedzai/commons/sql/abstraction/engine/AbstractDatabaseEngine.java
```diff
@@ -735,6 +736,27 @@ private void dropAllEntities() {
...
+            drop = conn.createStatement();
+            final String query = format("DROP VIEW %s", quotize(view, escapeCharacter()));
+            logger.trace(query);
+            drop.executeUpdate(query);
+        } catch (final SQLException ex) {
+            throw new DatabaseEngineException("Error dropping view", ex);
+        } finally {
+            try {
+                if (drop != null) {
+                    drop.close();
```
dfialho, line 752: try-with-resources is a cleaner and safer way to close resources

src/test/java/com/feedzai/commons/sql/abstraction/engine/impl/abs/EngineGeneralTest.java
```diff
@@ -1811,17 +1811,19 @@ public void joinThreeQueriesTest() throws DatabaseEngineException {
...
     // https://github.com/cockroachdb/cockroach/issues/10028
-    public void createViewTest() throws DatabaseEngineException {
+    public void createAndDropViewTest() throws DatabaseEngineException {
         test5Columns();
 
         try {
-            engine.executeUpdate("DROP VIEW " + quotize("VN", engine.escapeCharacter()));
+            engine.dropView("VN");
         } catch (final Throwable a) {
         }
```
dfialho, line 1820: I know this was already like this, but it is weird we are dropping the view before creating. If the goal is to assert that it throws an exception, then we should assert on it, otherwise this is not doing anything.

README.md
```diff
@@ -815,6 +815,17 @@ engine.executeUpdate(view);
...
+
+After creating a view, you may also want to delete it.
+
+```java
+engine.dropView("view_name");
+```
+|Function|Description|
+|:---|:---|
+|[dropView](https://feedzai.github.io/pdb/com/feedzai/commons/sql/abstraction/engine/AbstractDatabaseEngine.html#dropView(java.lang.String))|Drops a view with the given name.|
+
```
dfialho, line 828: :star: for adding the documentation

src/main/java/com/feedzai/commons/sql/abstraction/engine/DatabaseEngine.java
```diff
@@ -132,6 +132,16 @@
...
 
+    /**
+     * Drops a view.
+     *
+     * @param view The view name.
+     * @throws DatabaseEngineException If something goes wrong while dropping the view.
+     *
+     * @since 2.5.3
+     */
+    void dropView(final String view) throws DatabaseEngineException;
```
marpereira, line 143: Hello @nuno-santos96 (cc @dfialho and @joaonuno),
This is a breaking change, it will break the code when extending DatabaseEngine class.

PR #142, opened 2020-06-22
Fix javadoc plugin configuration and remove default implementation

jcsf:
The javadoc plugin configuration was wrong for the version 3.0.0, this commit fixes it.
Also the default implementation of the new interface method DatabaseEngine#dropView was removed since next release will be a new minor.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -327,19 +327,6 @@
             </plugin>
             <plugin>
                 <groupId>org.apache.maven.plugins</groupId>
-                <artifactId>maven-gpg-plugin</artifactId>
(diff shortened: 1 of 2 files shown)


pom.xml
```diff
@@ -326,24 +326,28 @@
                 </configuration>
             </plugin>
             <plugin>
-                <groupId>org.apache.maven.plugins</groupId>
```
TiagoMSSantos, line 329: Don't we need the groupId?

jcsf: It was already missing it. But I will add it

PR #145, opened 2020-07-08
Lock down plugin versions set by parent POM on release

bernardomaciel:
**Summary**

This patch forces the plugin versions to the ones desired
by setting the version on the `plugins` section.

When using the release plugin, the `sonatype-oss-release` profile
from the Sonatype OSS Parent POM is activated.

This means that the versions of some plugins (e.g. the
`maven-javadoc-plugin`) are set to different versions than
those specified on the PDB root POM `pluginManagement` section.

Additionally,

* the version of the parent POM was bumped to 9
* the POM file was cleaned up

**Test Plan**

`mvn help:effective-pom | grep -A 3 -E "maven-(javadoc)|(source)|(gpg)-plugin"`

With the `deploy` phase changed to `package` on the affected plugins:

`mvn -Prelease.signature -Darguments=-Dgpg.passphrase=thephrase release:prepare`
`mvn -Prelease.signature -DdryRun=true release:perform`

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -354,9 +354,7 @@
                 <artifactId>maven-javadoc-plugin</artifactId>
                 <version>${maven-javadoc-plugin.version}</version>
                 <configuration>
-                    <additionalOptions>
-                        <additionalOption>-Xdoclint:none</additionalOption>
-                    </additionalOptions>
+                    <doclint>none</doclint>
                 </configuration>
                 <executions>
                     <execution>
--- a/pom.xml
+++ b/pom.xml
@@ -1,5 +1,6 @@
 <?xml version="1.0"?>
-<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+<project xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xmlns="http://maven.apache.org/POM/4.0.0"
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
     <modelVersion>4.0.0</modelVersion>
     <groupId>com.feedzai</groupId>
     <artifactId>pdb</artifactId>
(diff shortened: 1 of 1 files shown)


pom.xml
```diff
@@ -130,7 +135,7 @@
         <dependency>
             <groupId>mysql</groupId>
             <artifactId>mysql-connector-java</artifactId>
-           <scope>provided</scope>
+            <scope>provided</scope>
```
jmf-tls, line 138: no need for the extra space :smile:

pom.xml
```diff
@@ -341,6 +352,7 @@
             <plugin>
                 <groupId>org.apache.maven.plugins</groupId>
                 <artifactId>maven-javadoc-plugin</artifactId>
+                <version>${maven-javadoc-plugin.version}</version>
                 <configuration>
                     <additionalOptions>
                         <additionalOption>-Xdoclint:none</additionalOption>
```
jmf-tls, line 358: I see that this was updated on a recent commit (it had `<additionalparam>` before).

If I had to guess, I'd say that change fixed javadoc generation locally, but broke it in on sonatype,
and this commit updates the maven-javadoc-plugin so that it will work again?

Anyway, if we're using the plugin v3.0.0, there is a specific tag for this.
Instead of passing options to the javadoc tool explicitly, we can simply use the _doclint_  tag
```suggestion
                    <doclint>none</doclint>
```
(I couldn't include the `</additionalOptions>` line in the suggestion, but if you accept it, then that line has to be removed too)

pom.xml
jmf-tls: I see now...
This was actually misaligned, nice

pom.xml
bernardomaciel: > If I had to guess, I'd say that change fixed javadoc generation locally, but broke it in on sonatype,
and this commit updates the maven-javadoc-plugin so that it will work again?
That's correct.

> Anyway, if we're using the plugin v3.0.0, there is a specific tag for this.
> Instead of passing options to the javadoc tool explicitly, we can simply use the doclint tag
Nice! Changed accordingly.

bernardomaciel:
I accidentally did "Rebase and merge" instead of squashing.
I rebased `master` interactively and pushed forcefully so that I could squash the two commits of this PR in one.

cc @jcsf @jmf-tls

PR #149, opened 2020-08-13
Add current row count information to ResultIterator

helenapoleri:
These changes introduce a way to identify the current row count (number of rows processed) of the ResultSet being iterated in the ResultIterator. A simple integer was used to avoid having to resort to ResultSet cursors.


src/main/java/com/feedzai/commons/sql/abstraction/dml/result/ResultIterator.java
```diff
@@ -126,6 +132,15 @@ public ResultIterator(PreparedStatement statement) throws DatabaseEngineExceptio
         this(statement, null, true);
     }
 
+    /**
+     * Retrieves the value for the index of the row being currently processed by the iterator.
+     *
+     * @return The value for the index of the row being currently processed by the iterator.
```
lmtjalves, line 138: - Add a note that if the ResultIterator is closed, then this will return 0.
- Also, make it clear that by current row index, you mean the index of the next row that will be returned by the next()/nextResult() method.

src/main/java/com/feedzai/commons/sql/abstraction/dml/result/ResultIterator.java
```diff
@@ -167,8 +186,12 @@ public ResultIterator(PreparedStatement statement) throws DatabaseEngineExceptio
         }
     }
 
+    public ResultSet getResultSet() {
```
hugomiguelabreu, line 189: javadoc missing.

src/test/java/com/feedzai/commons/sql/abstraction/dml/result/ResultIteratorTest.java
```diff
@@ -0,0 +1,57 @@
...
+ * @since 2.7.1
+ */
+public class ResultIteratorTest {
+
+    @Test
+    public void doesRowCountIncrementTest(@Mocked Statement statement,
+                                          @Mocked ResultSet resultSet)
+            throws DatabaseEngineException, SQLException {
+
+        MySqlResultIterator resultIterator = new MySqlResultIterator(statement, null);
```
hugomiguelabreu, line 27: Use the `ResultIterator` on the variable declaration there's no need to be Mysql specific.
Also the variable can be final.

src/test/java/com/feedzai/commons/sql/abstraction/dml/result/ResultIteratorTest.java
```diff
@@ -0,0 +1,57 @@
...
+/**
+ * Class for testing the result iterator behaviour.
+ *
+ * @author Helena Poleri (helena.poleri@feedzai.com)
+ * @since 2.7.1
+ */
+public class ResultIteratorTest {
+
+    @Test
+    public void doesRowCountIncrementTest(@Mocked Statement statement,
```
hugomiguelabreu, line 23: javadoc missing

src/main/java/com/feedzai/commons/sql/abstraction/dml/result/ResultIterator.java
helenapoleri: Since index was sort of misleading, I changed "current row index" to "current row counter", and updated the documentation. I also added a unit test for this.

src/test/java/com/feedzai/commons/sql/abstraction/dml/result/ResultIteratorTest.java
```diff
@@ -0,0 +1,57 @@
...
+
+import static org.junit.Assert.assertEquals;
+
+/**
+ * Class for testing the result iterator behaviour.
+ *
+ * @author Helena Poleri (helena.poleri@feedzai.com)
+ * @since 2.7.1
+ */
+public class ResultIteratorTest {
```
hugomiguelabreu, line 20: (probably for this comment is overkill) but, you could create a AbstractResultIteratorTest and then one for each engine that would provide the correct result iterator for the test.
I ok with this since for a simple test I don't think we should have that overhead but calling out @marpereira to check that.

src/main/java/com/feedzai/commons/sql/abstraction/dml/result/ResultIterator.java
helenapoleri: Removed this method, since it's not really necessary.

src/test/java/com/feedzai/commons/sql/abstraction/dml/result/ResultIteratorTest.java
marpereira: discussed offline

PR #150, opened 2020-10-20
Use native connection checks where possible and with configurable timout

hugomiguelabreu:
This commit introduces changes to the verification of the JDBC connections.
When verifying database connections, PDB should do it in a timely manner to allow a faster failover of the database. To achieve this goal, a new property is included to configure the timeout of the verification.
Most of the engines are now using JDBC standard `isValid()`. 
PostgreSQL, do not respect it and unit tests had failed consistently. (see: https://github.com/pgjdbc/pgjdbc/pull/1943)
DB2 does not correctly recover connections using `isValid()`.

### DIFF
--- a/src/main/java/com/feedzai/commons/sql/abstraction/engine/AbstractDatabaseEngine.java
+++ b/src/main/java/com/feedzai/commons/sql/abstraction/engine/AbstractDatabaseEngine.java
@@ -165,7 +165,7 @@ public abstract class AbstractDatabaseEngine implements DatabaseEngine {
      * An {@link ExecutorService} to be used by the DB drivers to break a connection if it has been blocked for longer
      * than the specified socket timeout
      */
-    private final ExecutorService socketTimeoutExecutor = Executors.newSingleThreadExecutor();
+    protected final ExecutorService socketTimeoutExecutor = Executors.newSingleThreadExecutor();
 
     /**
      * The default instance of {@link QueryExceptionHandler} to be used in disambiguating SQL exceptions.
@@ -1101,7 +1101,14 @@ public abstract class AbstractDatabaseEngine implements DatabaseEngine {
      * @param conn The connection to test.
      * @return True if the connection is valid, false otherwise.
      */
-    protected abstract boolean checkConnection(final Connection conn);
+    protected boolean checkConnection(final Connection conn) {
+        try {
+            return conn.isValid(this.properties.getCheckConnectionTimeout());
+        } catch (final Exception ex) {
+            logger.debug("Connection is down.", ex);
+            return false;
+        }
+    };
 
     /**
      * Checks if the connection is alive.
--- a/src/main/java/com/feedzai/commons/sql/abstraction/engine/impl/DB2Engine.java
+++ b/src/main/java/com/feedzai/commons/sql/abstraction/engine/impl/DB2Engine.java
@@ -803,6 +803,43 @@ public class DB2Engine extends AbstractDatabaseEngine {
 
     @Override
     protected boolean checkConnection(final Connection conn) {
+        final int timeout = this.properties.getCheckConnectionTimeout();
+        final int socketTimeout;
+        try {
+            socketTimeout = conn.getNetworkTimeout();
+            try {
+                // Set the socket timeout to verify the connection.
+                conn.setNetworkTimeout(socketTimeoutExecutor, timeout * 1000);
+                return pingConnection(conn);
+            } catch (final Exception ex) {
+                logger.debug("It wasn't possible to verify the connection state within the timeout of {} seconds.",
+                        timeout,
+                        ex);
+                return false;
+            } finally {
+                // Make sure to respect it afterwards.
+                conn.setNetworkTimeout(socketTimeoutExecutor, socketTimeout);
+            }
+        } catch (final Exception ex) {
+            logger.warn("It wasn't possible to reset the connection / fetch the timeout.");
+
+            try {
+                conn.close();
+            } catch (final Exception e) {
+                logger.debug("Error closing the connection.", e);
+            }
+
+            return false;
+        }
+    }
+
+    /**
+     * Executes a dummy query to verify if the connection is alive.
+     *
+     * @param conn The connection to test.
+     * @return {@code true} if the connection is valid, {@code false} otherwise.
+     */
+    private boolean pingConnection(final Connection conn) {
         Statement s = null;
         try {
             s = conn.createStatement();
--- a/src/main/java/com/feedzai/commons/sql/abstraction/engine/impl/MySqlEngine.java
+++ b/src/main/java/com/feedzai/commons/sql/abstraction/engine/impl/MySqlEngine.java
@@ -752,28 +752,6 @@ public class MySqlEngine extends AbstractDatabaseEngine {
     }
 
     @Override
-    protected boolean checkConnection(final Connection conn) {
-        Statement s = null;
-        try {
-            s = conn.createStatement();
-            s.executeQuery("select 1");
-
-            return true;
-        } catch (final SQLException e) {
-            logger.debug("Connection is down.", e);
-            return false;
-        } finally {
-            try {
-                if (s != null) {
-                    s.close();
-                }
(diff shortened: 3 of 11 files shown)


hugomiguelabreu:
Unit tests are good to go.

src/main/java/com/feedzai/commons/sql/abstraction/engine/AbstractDatabaseEngine.java
```diff
@@ -1103,6 +1104,50 @@ private void validateEntity(final DbEntity entity) throws DatabaseEngineExceptio
...
+                return pingConnection(conn);
+            } catch (final Exception ex) {
+                logger.debug("Connection is down.", ex);
+                return false;
+            } finally {
+                // Make sure to respect it afterwards.
+                conn.setNetworkTimeout(socketTimeoutExecutor, socketTimeout);
+            }
+        } catch (final Exception ex) {
+            logger.debug("Connection is down.", ex);
```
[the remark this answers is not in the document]
artpdr: I would also try to close the connection here just in case `conn.close()`. In case the previous exception didn't set the timeout to the old one but the connection wasn't also closed

pom.xml
```diff
@@ -59,7 +59,7 @@
             <dependency>
                 <groupId>org.postgresql</groupId>
                 <artifactId>postgresql</artifactId>
-                <version>42.2.14</version>
+                <version>42.2.18</version>
```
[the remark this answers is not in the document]
hugomiguelabreu: Well since PostgreSQL patch 14 has a problem with socket timeouts and we are using them on this review I think it goes well with it.
WDYT?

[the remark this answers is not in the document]
artpdr: I agree!

hugomiguelabreu:
UT are all good to go.

src/main/java/com/feedzai/commons/sql/abstraction/engine/impl/MySqlEngine.java
```diff
@@ -753,6 +753,11 @@ protected void dropReferringFks(DbEntity entity) throws DatabaseEngineException
 
     @Override
     protected boolean checkConnection(final Connection conn) {
+        return checkConnection(conn, this.properties.getCheckConnectionTimeout());
+    }
+
+    @Override
+    protected boolean pingConnection(final Connection conn) {
```
artpdr, line 760: WDYT of making pingConnection a default method in the interface with the most common implementation and just change in the implementations where it is different (I think it's only on db2). 

Due to your findings, we don't need a pingConnection on MySQL and Oracle. Did you confirm that we don't suffer from the same problem in the other implementations. Nice job finding this, though

hugomiguelabreu: Oracle is also different, but yeah you got a good point there! On it.

src/main/java/com/feedzai/commons/sql/abstraction/engine/impl/OracleEngine.java
```diff
@@ -1035,6 +1035,11 @@ protected void addFks(DbEntity entity) throws DatabaseEngineException {
 
     @Override
     protected boolean checkConnection(final Connection conn) {
+        return checkConnection(conn, this.properties.getCheckConnectionTimeout());
+    }
+
+    @Override
+    protected boolean pingConnection(final Connection conn) {
```
artpdr, line 1042: I think this isn't needed for oracle since we are always entering in the checkconnection below that used the conn.isValid(timeout). Am I wrong?

src/main/java/com/feedzai/commons/sql/abstraction/engine/AbstractDatabaseEngine.java
```diff
@@ -1103,6 +1104,88 @@ private void validateEntity(final DbEntity entity) throws DatabaseEngineExceptio
...
+     *     <li>DB2</li>
+     * </ul>
+     * PostgreSQL is not respecting the timeout in the {@link Connection#isValid(int)}.
+     *
+     * @param conn    The connection to test.
+     * @param timeout The timeout in seconds.
+     * @return {@code true} if the connection is valid, {@code false} otherwise.
+     * @since 2.7.2
+     */
+    protected boolean checkConnection(final Connection conn, final int timeout) {
```
artpdr, line 1159: Sorry but if this checkConnection is intended to be used only in the Postgres implementation, why don't we move this method to that specific implementation and move here the method that does the #isValid check (we can then remove the checkConnection from the other implementations).

I would also make the #pingConnection visible only in the Postgres class since it just needs to be used on that implementation.

Do you agree?

artpdr (approved):
Nice job!

hugomiguelabreu:
DB2 using `isValid()` isn't fully supported on PDB. I left the original implementation. This requires a re-review when possible @artpdr

src/main/java/com/feedzai/commons/sql/abstraction/engine/impl/DB2Engine.java
```diff
@@ -803,6 +803,12 @@ protected void addFks(DbEntity entity) throws DatabaseEngineException {
 
     @Override
     protected boolean checkConnection(final Connection conn) {
+        return checkConnection(conn, this.properties.getCheckConnectionTimeout());
+    }
+
+
+    @Override
+    protected boolean pingConnection(final Connection conn) {
```
artpdr, line 811: Why do we have this method here? Can't we have this like `PostgreSqlEngine`?

artpdr: Nvm... they need different queries!

src/main/java/com/feedzai/commons/sql/abstraction/engine/AbstractDatabaseEngine.java
```diff
@@ -71,6 +71,7 @@
 import java.util.Optional;
 import java.util.Properties;
 import java.util.Set;
+import java.util.concurrent.Executor;
```
jmf-tls, line 74: This is unused, remove
```suggestion
```

src/test/java/com/feedzai/commons/sql/abstraction/engine/impl/abs/TimeoutsTest.java
```diff
@@ -164,11 +172,28 @@ public void testTimeoutsConfigured() throws Exception {
...
-        int connectionTimeoutInMs = de.getConnection().getNetworkTimeout();
+        int socketConnectionTimeoutInMs = de.getConnection().getNetworkTimeout();
+        int checkConnectionTimeoutInMs = de.getProperties().getCheckConnectionTimeout();
 
         assertEquals("Is the login timeout of the DB connection the expected?", LOGIN_TIMEOUT_SECONDS, loginTimeoutInSeconds);
         assertEquals("Is the socket timeout of the DB connection the expected?",
-                TimeUnit.SECONDS.toMillis(SOCKET_TIMEOUT_SECONDS), connectionTimeoutInMs);
+                TimeUnit.SECONDS.toMillis(SOCKET_TIMEOUT_SECONDS), socketConnectionTimeoutInMs);
+        assertEquals("Is the check connection timeout of the DB connection the expected?",
+                TimeUnit.SECONDS.toSeconds(CHECK_CONNECTION_TIMEOUT_SECONDS), checkConnectionTimeoutInMs);
```
jmf-tls, line 182: Connection timeout config is PDB stuff, it's not "in milliseconds", so `checkConnectionTimeoutInMs` should be named accordingly.
Also, as the name implies, CHECK_CONNECTION_TIMEOUT_SECONDS is already seconds, you don't need to convert seconds to seconds :smile: 
```suggestion
        assertEquals("Is the check connection timeout of the DB connection the expected?",
                CHECK_CONNECTION_TIMEOUT_SECONDS, checkConnectionTimeout);
```

jmf-tls (commented):
looks great, only minor stuff to check

jmf-tls (approved):
Please adjust the title and summary to what's currently being done

PR #156, opened 2020-10-22
Create CI for build PDB

hugomiguelabreu:


.github/workflows/build-pdb.yml
```diff
@@ -0,0 +1,24 @@
...
+    runs-on: ubuntu-latest
+
+    steps:
+    - uses: actions/checkout@v2
+    - name: Set up JDK 1.8
+      uses: actions/setup-java@v1
+      with:
+        java-version: 1.8
+    - name: Build with Maven
+      run: mvn -B package --file pom.xml -DskipTests
```
TiagoMSSantos, line 24: Shouldn't we do `mvn clean install` first?

hugomiguelabreu: No. Maven package is exactly the same as package, but install puts the binaries into your .m2. That's useless here since we wont be reusing it for now.

artpdr, line 24: good stuff... Shouldn't we let tests run?

hugomiguelabreu: See here: https://stackoverflow.com/questions/16602017/how-are-mvn-clean-package-and-mvn-clean-install-different

hugomiguelabreu: Glad you asked :smile: 
I separated that into more actions to have it segregated: https://github.com/feedzai/pdb/pull/157

hugomiguelabreu: See in example how Pulse separates things: https://github.com/apache/pulsar/actions

marpereira (approved):

.github/workflows/build-pdb.yml
TiagoMSSantos: And with `package` it will also run the tests right?

TiagoMSSantos: But won't the new workflow to run the tests also have to compile the code?

hugomiguelabreu: Yes they do have to compile the code. It's not a big deal, compiling PDB is 4 seconds. We can use caching also. But I'll have to check how it works inter-workflow.

hugomiguelabreu: @TiagoMSSantos This is now using a cache. Let's see how it'll behave in the future workflows.

TiagoMSSantos: Ok then, it looks good to me.

TiagoMSSantos (approved):
Looks good to me!
Nice one about the cache by the way.

artpdr (approved):

PR #158, opened 2020-10-22
CI tests for PostgreSQL

hugomiguelabreu:


.github/workflows/test-postgresql.yml
```diff
@@ -0,0 +1,31 @@
...
+
+    - name: Cache Maven packages
+      uses: actions/cache@v2
+      with:
+        path: ~/.m2
+        key: ${{ runner.os }}-m2-${{ hashFiles('**/pom.xml') }}
+        restore-keys: ${{ runner.os }}-m2
+
+    - name: Test with PostgreSQL
+      run: mvn -B clean test -Ppostgresql
```
artpdr, line 31: Question: On average, how much time will tests take in each job?

hugomiguelabreu: A few minutes. PostgreSQL is 3 minutes. But DB2 locally takes 10 minutes.

artpdr: I was thinking in running tests for each profile but 10 minutes is probably too much

artpdr (approved):

.github/workflows/test-postgresql.yml
hugomiguelabreu: We will :) but each profile can go into a single workflow, this way they can run in parallel

PR #160, opened 2020-10-23
Create tests-mysql.yml

hugomiguelabreu:


src/test/java/com/feedzai/commons/sql/abstraction/engine/impl/abs/TimeoutsTest.java
```diff
@@ -170,16 +180,19 @@ public void cleanResources() {
...
+        final Connection connection = de.getConnection();
 
-        int loginTimeoutInSeconds = DriverManager.getLoginTimeout();
-        int socketConnectionTimeoutInMs = de.getConnection().getNetworkTimeout();
-        int checkConnectionTimeoutInSec = de.getProperties().getCheckConnectionTimeout();
+        // Some engines (e.g. mysql) set the network timeout asynchronously.
+        await().atMost(Duration.ofSeconds(5))
+               .until(() ->
+                   TimeUnit.SECONDS.toMillis(SOCKET_TIMEOUT_SECONDS) == connection.getNetworkTimeout()
+               );
```
jmf-tls, line 189: This results in something like
`org.awaitility.core.ConditionTimeoutException: Condition with com.feedzai.commons.sql.abstraction.engine.impl.abs.TimeoutsTest was not fulfilled within 5 seconds.`

We should better identify this failure in the logs, and we can do the assertion right here:
```suggestion
        await("await network timeout set")
                .atMost(Duration.ofSeconds(5))
                .untilAsserted(() -> assertEquals("Is the socket timeout of the DB connection the expected?",
                        TimeUnit.SECONDS.toMillis(SOCKET_TIMEOUT_SECONDS), socketConnectionTimeoutInMs)
                );
```
This results in
```
org.awaitility.core.ConditionTimeoutException: Condition with alias 'await network timeout set' didn't complete within 5 seconds because assertion condition defined as a com.feedzai.commons.sql.abstraction.engine.impl.abs.TimeoutsTest Is the socket timeout of the DB connection the expected? expected:<20> but was:<0>.

........... stacktrace of the waiter .............

Caused by: java.lang.AssertionError: Is the socket timeout of the DB connection the expected? expected:<20> but was:<0>

........... stacktrace of the assert .............
```

src/test/java/com/feedzai/commons/sql/abstraction/engine/impl/abs/TimeoutsTest.java
```diff
@@ -170,16 +180,19 @@ public void cleanResources() {
     @Test
     public void testTimeoutsConfigured() throws Exception {
         final DatabaseEngine de = DatabaseFactory.getConnection(this.dbProps);
+        final Connection connection = de.getConnection();
 
-        int loginTimeoutInSeconds = DriverManager.getLoginTimeout();
-        int socketConnectionTimeoutInMs = de.getConnection().getNetworkTimeout();
-        int checkConnectionTimeoutInSec = de.getProperties().getCheckConnectionTimeout();
```
jmf-tls, line 176: I know these are kind of useless, but I'd rather keep them.
I think they make it more clear that some timeouts are seconds and others ms,
instead of having to infer those time units by looking at the expected constants in the assertions.

jmf-tls (approved):

TiagoMSSantos (approved):

PR #164, opened 2020-10-28
H2Engine: Verify if the table has an auto increment column before trying to read the generated value

TiagoMSSantos:
This commit fixes the H2 Engine when trying to persist a given entry in the database.

When persisting an entry with auto increment as `true`, it is returned the auto generated value from the generated keys, but in the newest version of H2 1.4.200, the method `PreparedStatement#getGeneratedKeys` returns all the columns as containing auto generated values by default, so it is necessary check if the that table has an auto increment column.

### DIFF
--- a/src/main/java/com/feedzai/commons/sql/abstraction/engine/impl/H2Engine.java
+++ b/src/main/java/com/feedzai/commons/sql/abstraction/engine/impl/H2Engine.java
@@ -375,7 +375,7 @@ public class H2Engine extends AbstractDatabaseEngine {
 
     @Override
     protected MappedEntity createPreparedStatementForInserts(final DbEntity entity) throws DatabaseEngineException {
-        List<String> insertInto = new ArrayList<>();
+        final List<String> insertInto = new ArrayList<>();
         insertInto.add("INSERT INTO");
         insertInto.add(quotize(entity.getName()));
         final List<String> insertIntoWithAutoInc = new ArrayList<>();
@@ -385,10 +385,14 @@ public class H2Engine extends AbstractDatabaseEngine {
         final List<String> values = new ArrayList<>();
         final List<String> columnsWithAutoInc = new ArrayList<>();
         final List<String> valuesWithAutoInc = new ArrayList<>();
+        String columnWithAutoIncName = null;
         for (final DbColumn column : entity.getColumns()) {
             columnsWithAutoInc.add(quotize(column.getName()));
             valuesWithAutoInc.add("?");
-            if (!column.isAutoInc()) {
+            if (column.isAutoInc()) {
+                // Get the only column name with auto incremented values.
+                columnWithAutoIncName = column.getName();
+            } else {
                 columns.add(quotize(column.getName()));
                 values.add("?");
             }
@@ -412,11 +416,19 @@ public class H2Engine extends AbstractDatabaseEngine {
         final PreparedStatement ps, psReturn, psWithAutoInc;
         try {
 
-            ps = conn.prepareStatement(statement, Statement.RETURN_GENERATED_KEYS);
-            psReturn = conn.prepareStatement(insertReturnStatement);
-            psWithAutoInc = conn.prepareStatement(statementWithAutoInt);
+            // Generate keys when the table has at least 1 column with auto generate value.
+            final int generateKeys = columnWithAutoIncName != null? Statement.RETURN_GENERATED_KEYS : Statement.NO_GENERATED_KEYS;
+            ps = this.conn.prepareStatement(statement, generateKeys);
+            psReturn = this.conn.prepareStatement(insertReturnStatement, generateKeys);
+            psWithAutoInc = this.conn.prepareStatement(statementWithAutoInt, generateKeys);
 
-            return new MappedEntity().setInsert(ps).setInsertReturning(psReturn).setInsertWithAutoInc(psWithAutoInc);
+            return new MappedEntity()
+                .setInsert(ps)
+                .setInsertReturning(psReturn)
+                .setInsertWithAutoInc(psWithAutoInc)
+                // The auto incremented column must be set, so when persisting a row, its possible to retrieve its value
+                // by consulting the column name from this MappedEntity.
+                .setAutoIncColumn(columnWithAutoIncName);
         } catch (final SQLException ex) {
             throw new DatabaseEngineException("Something went wrong handling statement", ex);
         }
@@ -533,13 +545,17 @@ public class H2Engine extends AbstractDatabaseEngine {
     protected synchronized long doPersist(final PreparedStatement ps,
                                           final MappedEntity me,
                                           final boolean useAutoInc,
-                                          int lastBindPosition) throws Exception {
+                                          final int lastBindPosition) throws Exception {
         ps.execute();
 
-        if (useAutoInc) {
+        final String autoIncColumnName = me.getAutoIncColumn();
+        if (useAutoInc && autoIncColumnName != null) {
             try (final ResultSet generatedKeys = ps.getGeneratedKeys()) {
                 if (generatedKeys.next()) {
-                    return generatedKeys.getLong(1);
+                    // These generated keys belong to the primary key and might not have auto incremented values nor
+                    // even have integer values.
+                    // That's why its necessary to only retrieve the value from a column that is auto incremented.
+                    return generatedKeys.getLong(autoIncColumnName);
                 }
             }
         }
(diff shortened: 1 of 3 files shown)


src/main/java/com/feedzai/commons/sql/abstraction/engine/impl/H2Engine.java
```diff
@@ -533,13 +533,17 @@ protected PreparedStatement getPreparedStatementForPersist(final boolean useAuto
...
-            try (final ResultSet generatedKeys = ps.getGeneratedKeys()) {
-                if (generatedKeys.next()) {
-                    return generatedKeys.getLong(1);
+            for (final DbColumn column : me.getEntity().getColumns()) {
+                if ((column.isAutoInc())) {
+                    try (final ResultSet generatedKeys = ps.getGeneratedKeys()) {
+                        if (generatedKeys.next()) {
+                            return generatedKeys.getLong(column.getName());
+                        }
+                    }
```
marpereira, line 546: I don't know the performance of a cycle for versus something like `final DbColumn column = me.getEntity().getColumns().stream().filter(c -> c.isAutoInc()).findFirst().get();`.
This is an important method that we need to try not to introduce cyclomatic complexity.

marpereira: Actually, MappedEntity has `.getAutoIncColumn()`.
You just need to change the `return generatedKeys.getLong(1);` to `return generatedKeys.getLong(me.getAutoIncColumn());`

hugomiguelabreu: For H2 `getAutoIncColumn()` is not being set :disappointed: 
We can either find a wait to set it on the binding of the statement on `entityToPreparedStatement()` otherwise there's always the need to go through the columns.

I agree in the stream by the way :smile:

artpdr: Probably we need to set the column as auto increment when creating it... https://www.baeldung.com/jdbc-returning-generated-keys

marpereira (approved):
LGTM

src/main/java/com/feedzai/commons/sql/abstraction/engine/impl/H2Engine.java
```diff
@@ -385,10 +385,13 @@ protected MappedEntity createPreparedStatementForInserts(final DbEntity entity)
         final List<String> values = new ArrayList<>();
         final List<String> columnsWithAutoInc = new ArrayList<>();
         final List<String> valuesWithAutoInc = new ArrayList<>();
+        String columnWithAutoIncName = null;
         for (final DbColumn column : entity.getColumns()) {
             columnsWithAutoInc.add(quotize(column.getName()));
             valuesWithAutoInc.add("?");
-            if (!column.isAutoInc()) {
+            if (column.isAutoInc()) {
```
artpdr, line 392: What if we have more than 1 auto_increment column? Shouldn't we save these names in a list and in the #doPersist use for example the first name that appears on the list in position 0? We need to specify in the documentation that we assume that the first column with auto_increment is the one that contains the entity ID

src/main/java/com/feedzai/commons/sql/abstraction/engine/impl/H2Engine.java
TiagoMSSantos: This new fix already creates a new `MappedEntity` with the last column found that contains `autoinc` values set as the `IncColumn` in the `MappedEntity`.

src/main/java/com/feedzai/commons/sql/abstraction/engine/impl/H2Engine.java
artpdr: If we have more than one column as auto_increment we will end up assuming that the one that provides entity id is the last one. We can also continue using a string instead of a list but don't set this string more than once.

TiagoMSSantos: I think we should keep using just a simple `String` because if we added the functionality to retrieve every auto incremented value,  it would break the current API, right?
So I think the best approach is either return the auto incremented value from the first or last column and state it in the Javadoc.

artpdr: I agree with having just a string since we just want to retrieve one auto incremented column as the entity id. I think it makes sense to be the first auto incremented column that we observe. Normally primary keys are defined first in CREATE TABLE statements

src/main/java/com/feedzai/commons/sql/abstraction/engine/impl/H2Engine.java
```diff
@@ -385,10 +385,14 @@ protected MappedEntity createPreparedStatementForInserts(final DbEntity entity)
         final List<String> values = new ArrayList<>();
         final List<String> columnsWithAutoInc = new ArrayList<>();
         final List<String> valuesWithAutoInc = new ArrayList<>();
+        String columnWithAutoIncName = null;
         for (final DbColumn column : entity.getColumns()) {
             columnsWithAutoInc.add(quotize(column.getName()));
```
artpdr, line 390: Just noticed now... but why do we add all columns to this list, even the ones that aren't auto_increment?

src/main/java/com/feedzai/commons/sql/abstraction/engine/impl/H2Engine.java
```diff
@@ -385,10 +385,14 @@ protected MappedEntity createPreparedStatementForInserts(final DbEntity entity)
...
             valuesWithAutoInc.add("?");
-            if (!column.isAutoInc()) {
+            // Find the name of the first column that contains an auto generated value.
```
marpereira, line 392: not first, but the last

TiagoMSSantos: It is now the first one.

src/main/java/com/feedzai/commons/sql/abstraction/engine/impl/H2Engine.java
TiagoMSSantos: I know its weird, but not adding all the columns to that list, make some tests fail.

hugomiguelabreu:
Overall looks good to go. Just retrigger the SQLServer action.

hugomiguelabreu (approved):

src/main/java/com/feedzai/commons/sql/abstraction/engine/impl/H2Engine.java
hugomiguelabreu: Have you analyzed the test so see if they need to be adapted is this expected?

TiagoMSSantos: That's what I'm trying to understand right now.

artpdr: Yes it's strange reading this code and watching it adding all the columns to a list named columnWithAutoIncName... I don't know this code but please check if this is ok. Maybe there is some place where it's being done a difference between the normal columns and columnsWithAutoInc in order to create a correct prepared statement :man_shrugging:

src/test/java/com/feedzai/commons/sql/abstraction/engine/impl/abs/EngineGeneralTest.java
```diff
@@ -1548,25 +1551,32 @@ public void getGeneratedKeysFromAutoIncWithTransactionTest() throws DatabaseEngi
...
+
+        final EntityEntry ee = entry()
+                .set("COL1", "VAL1")
+                .set("COL2", "VAL2")
                 .build();
 
-        Long persist = engine.persist("TEST", ee);
-
-        assertNull("ret null?", persist);
+        Assertions.assertThat(this.engine.persist("TEST", ee))
```
hugomiguelabreu, line 1577: Import Assertions and you can call the static assertThat()

src/main/java/com/feedzai/commons/sql/abstraction/engine/impl/H2Engine.java
```diff
@@ -385,10 +385,16 @@ protected MappedEntity createPreparedStatementForInserts(final DbEntity entity)
         final List<String> values = new ArrayList<>();
         final List<String> columnsWithAutoInc = new ArrayList<>();
         final List<String> valuesWithAutoInc = new ArrayList<>();
+        String columnWithAutoIncName = null;
         for (final DbColumn column : entity.getColumns()) {
             columnsWithAutoInc.add(quotize(column.getName()));
             valuesWithAutoInc.add("?");
-            if (!column.isAutoInc()) {
+            if (column.isAutoInc()) {
+                if (columnWithAutoIncName == null) {
```
marpereira, line 393: you can join these conditions or even delete the second one because it's confirmed that a DBEntity can only have 1 AutoInc Column

src/main/java/com/feedzai/commons/sql/abstraction/engine/impl/H2Engine.java
TiagoMSSantos: Well I tried to figure that out, but it seems that when we only add the columns that contain auto incremented values in those lists `columnsWithAutoInc` and `valuesWithAutoInc`, the `MappedEntity` used when binding it to the prepared statement, it always needs to contain all the columns that an `EntityEntry` might have.

TiagoMSSantos: Otherwise, it can't bind the properly and a `SQLException` is thrown.

PR #166, opened 2020-10-30
dependabot: Add dependabot config file

TiagoMSSantos:
This commit adds a config file for the `dependabot`.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,28 @@
+version: 2
+updates:
+# Enable version updates for Maven
+- package-ecosystem: "maven"
+  # Look for a `pom.xml` in the `root` directory
+  directory: "/"
+  # Check for updates every day (weekdays)
+  schedule:
+    interval: "daily"
+    # Check for Maven updates at 4am UTC
+    time: "04:00"
+    timezone: "Europe/Lisbon"
+  open-pull-requests-limit: 3
+  # Raise all Maven pull requests with reviewers
+  #reviewers:
+  #  - "<someone>"
+  # Raise all Maven pull requests with an assignee
+  #assignees:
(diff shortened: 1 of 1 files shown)


TiagoMSSantos:
The documentation about the configuration is in here: https://docs.github.com/en/free-pro-team@latest/github/administering-a-repository/configuration-options-for-dependency-updates

.github/dependabot.yml
```diff
@@ -0,0 +1,28 @@
...
+- package-ecosystem: "maven"
+  # Look for a `pom.xml` in the `root` directory
+  directory: "/"
+  # Check for updates every day (weekdays)
+  schedule:
+    interval: "daily"
+    # Check for Maven updates at 4am UTC
+    time: "04:00"
+    timezone: "Europe/Lisbon"
+  open-pull-requests-limit: 1
```
hugomiguelabreu, line 13: Do we really want to limit this?

TiagoMSSantos: We have to put a limit, otherwise it will probably create many pull requests at once.
Maybe just 1 is too low, and we can raise it to like 3 or 4 at a time.

smmf: Yes, I agree with having a limit. Any small number (e.g. between 1 and 5) should be ok for now. And we can tweak it later.

hugomiguelabreu: Let's use 3 as a rule of thumb and manage accordingly later. WDYT?

smmf (approved):
LGTM

.github/dependabot.yml
TiagoMSSantos: Agreed

hugomiguelabreu (approved):

PR #170, opened 2020-10-31
dependabot: Add schedule updates for GitHub Actions

TiagoMSSantos:
Add schedule updates for GitHub workflow Actions.
This commit will make `dependabot` also search for updates in the GitHub Actions used, like the checkout and cache for example that is currently being used, so if updates for those actions appear, this will make dependabot open new pull requests to update those.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -26,3 +26,29 @@ updates:
   allow:
     # Allow both direct and indirect updates for all packages
     - dependency-type: "all"
+
+# Set update schedule for GitHub Actions
+- package-ecosystem: "github-actions"
+  directory: "/"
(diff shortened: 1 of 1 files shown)


hugomiguelabreu:
Can you improve the description?
Why this is needed?
How does it work?
What is going to be changed?

TiagoMSSantos:
> Can you improve the description?
> 
>     * Why this is needed?
> 
>     * How does it work?
> 
>     * What is going to be changed?

Comment addressed. Thanks for the feedback!

hugomiguelabreu (approved):

helenapoleri (approved):
lgtm

PR #177, opened 2020-11-20
pom: Bump Guava to version 25.1

TiagoMSSantos:
This commit upgrades `Guava` to version `25.1-jre`.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -99,7 +99,7 @@
             <dependency>
                 <groupId>com.google.guava</groupId>
                 <artifactId>guava</artifactId>
-                <version>18.0</version>
+                <version>25.1-jre</version>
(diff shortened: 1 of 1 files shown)


pom.xml
```diff
@@ -99,7 +99,7 @@
             <dependency>
                 <groupId>com.google.guava</groupId>
                 <artifactId>guava</artifactId>
-                <version>18.0</version>
+                <version>25.1-jre</version>
```
artpdr, line 102: Can you remind me why don't we go with version 30?

TiagoMSSantos: Because Hadoop version 2 is only compatible with Guava up to version 25.

hugomiguelabreu (approved):

PR #187, opened 2020-12-22
Support Concat operator

### LINKED ISSUE #184: Support Concat operator

francisco-polaco:
In this PR, we've implemented the concat operator.

Closes #184


src/test/java/com/feedzai/commons/sql/abstraction/engine/impl/abs/EngineGeneralTest.java
```diff
@@ -2293,6 +2294,36 @@ public void testCaseMultipleWhenElse() throws DatabaseEngineException {
...
+        engine.persist("TEST", entry().set("COL1", 3).set("COL5", "xpto")
+                .build());
+        engine.persist("TEST", entry().set("COL1", 4).set("COL5", "teste")
+                .build());
+        engine.persist("TEST", entry().set("COL1", 5).set("COL5", "pomme de terre")
+                .build());
+
+        final Query query =
+                select(
+                        concat(k("."),
```
joaonuno, line 2313: Maybe try other delimiters to make sure the result is consistent among implementations. I would suggest at least null, empty, a number and a delimiter coming from another column.

francisco-polaco: Some implementations do not support `NULL` or mixing data types, eg. delimiter is a number and columns are varchars.

Do you think we should throw an error in these cases or just let the DB return the error?

jmf-tls:
This PR mentions the concat operator, which implements concatenation with a delimiter.
In SQL there is a CONCAT operator, that performs concatenation _**without**_ a delimiter,
while there is also a CONCAT_WS operator that uses a delimiter.

So PDB concat will perform CONCAT_WS, which is confusing.
How about implementing both in PDB, with names matching the SQL operators?

src/main/java/com/feedzai/commons/sql/abstraction/engine/impl/CockroachDBTranslator.java
```diff
@@ -105,6 +107,18 @@ public String translate(final Cast cast) {
...
+        inject(concat.getDelimiter());
+        inject(concat.getExpressions());
+
+        final String concatWs = format(" || %s || ", concat.getDelimiter().translate());
+
+        return concat.getExpressions().stream()
+                     .map(Expression::translate)
+                     .collect(Collectors.joining(concatWs));
+    }
+
```
jmf-tls, line 121: Why?
CockroachDB should support CONCAT_WS just fine

francisco-polaco:
> This PR mentions the concat operator, which implements concatenation with a delimiter.
> In SQL there is a CONCAT operator, that performs concatenation _**without**_ a delimiter,
> while there is also a CONCAT_WS operator that uses a delimiter.
> 
> So PDB concat will perform CONCAT_WS, which is confusing.
> How about implementing both in PDB, with names matching the SQL operators?

Indeed we are doing CONCAT_WS, which is the only being used right now. I believe we can introduce the regular CONCAT, once we need it. In case someone really needs CONCAT, an empty delimiter can be used to achieve the same purpose.

src/main/java/com/feedzai/commons/sql/abstraction/engine/impl/CockroachDBTranslator.java
francisco-polaco: CockroachDB is having issues in the tests with CONCAT_WS.

francisco-polaco: Fixed!

src/main/java/com/feedzai/commons/sql/abstraction/dml/Concat.java
```diff
@@ -0,0 +1,81 @@
...
+ */
+package com.feedzai.commons.sql.abstraction.dml;
+
+import java.util.Collection;
+
+/**
+ * PDB Concat Operator.
+ *
+ * @author Francisco Santos (francisco.santos@feedzai.com)
+ * @since @@@feedzai.next.release@@@
```
jcsf, line 24: I'm not sure that PDB has the materialization for these tokens or if you should just hard code the version.

src/test/java/com/feedzai/commons/sql/abstraction/engine/impl/abs/EngineGeneralTest.java
```diff
@@ -2293,6 +2294,96 @@ public void testCaseMultipleWhenElse() throws DatabaseEngineException {
...
+                        .from(table("TEST"));
+
+        final List<Map<String, ResultColumn>> result = engine.query(query);
+
+        assertEquals("testetesteteste", result.get(0).get("concat").toString());
+        assertEquals("xptoxptoxpto", result.get(1).get("concat").toString());
+        assertEquals("xptoxptoxpto", result.get(2).get("concat").toString());
+        assertEquals("testetesteteste", result.get(3).get("concat").toString());
+        assertEquals("pomme de terrepomme de terrepomme de terre", result.get(4).get("concat").toString());
+    }
```
jcsf, line 2385: All tests seem very similar can you just make a single function and pass the differences as an argument?

francisco-polaco: While I agree they are similar the assertions are different. WDYT of refactoring everything except the assertions?

src/main/java/com/feedzai/commons/sql/abstraction/dml/Concat.java
francisco-polaco: Yeah I don't let's make it `2.8.1`

jcsf (approved):

jmf-tls:
I also tested H2.
Using dbeaver,
```
INSERT INTO PUBLIC.NEWTABLE (COLUMN1, COLUMN2) VALUES ('vals', concat_ws(', ', 'a', 'b', ' ', 'c'));
INSERT INTO PUBLIC.NEWTABLE (COLUMN1, COLUMN2) VALUES ('nul', concat_ws(', ', 'a', 'b', NULL, 'c'));
INSERT INTO PUBLIC.NEWTABLE (COLUMN1, COLUMN2) VALUES ('empty', concat_ws(', ', 'a', 'b', '', 'c'));
```
results in
```
a, b,  , c
a, b, c
a, b, c
```
_________________________
Using PDB,
```
        engine.persist("TEST", entry().set("COL2", "ola1").build());
        engine.persist("TEST", entry().set("COL2", "ola2").build());
        engine.persist("TEST", entry().set("COL2", "ola3").build());

        engine.executeUpdate(update(table("TEST"))
                .set(eq(column("COL3"), concat(k(", "), k("a"), k("b"), k(" "), k("c"))))
                .where(eq(column("COL2"), k("ola1")))
        );
        engine.executeUpdate(update(table("TEST"))
                .set(eq(column("COL3"), concat(k(", "), k("a"), k("b"), k(null), k("c"))))
                .where(eq(column("COL2"), k("ola1")))
        );
        engine.executeUpdate(update(table("TEST"))
                .set(eq(column("COL3"), concat(k(", "), k("a"), k("b"), k(""), k("c"))))
                .where(eq(column("COL2"), k("ola1")))
        );
```
results in
```
a, b, c
null
null
```
_________________________
Both cases are strange. I know that they are fixing this stuff in H2, but until the new driver is released
I think we should give H2 the same treatment as Oracle and DB2.

With this changed, and the special case for NULL addressed also in Oracle and H2, you could follow the suggestion from @joaonuno and add some test cases with NULL and empty string.
Regarding your comment there, I agree that we should not consider mixed types as supported. For now, we should only have a guarantee for string types...

jmf-tls (commented):
I also tested H2.
Using dbeaver,
```
INSERT INTO PUBLIC.NEWTABLE (COLUMN1, COLUMN2) VALUES ('vals', concat_ws(', ', 'a', 'b', ' ', 'c'));
INSERT INTO PUBLIC.NEWTABLE (COLUMN1, COLUMN2) VALUES ('nul', concat_ws(', ', 'a', 'b', NULL, 'c'));
INSERT INTO PUBLIC.NEWTABLE (COLUMN1, COLUMN2) VALUES ('empty', concat_ws(', ', 'a', 'b', '', 'c'));
```
results in
```
a, b,  , c
a, b, c
a, b, c
```
_________________________
Using PDB,
```
        engine.persist("TEST", entry().set("COL2", "ola1").build());
        engine.persist("TEST", entry().set("COL2", "ola2").build());
        engine.persist("TEST", entry().set("COL2", "ola3").build());

        engine.executeUpdate(update(table("TEST"))
                .set(eq(column("COL3"), concat(k(", "), k("a"), k("b"), k(" "), k("c"))))
                .where(eq(column("COL2"), k("ola1")))
        );
        engine.executeUpdate(update(table("TEST"))
                .set(eq(column("COL3"), concat(k(", "), k("a"), k("b"), k(null), k("c"))))
                .where(eq(column("COL2"), k("ola1")))
        );
        engine.executeUpdate(update(table("TEST"))
                .set(eq(column("COL3"), concat(k(", "), k("a"), k("b"), k(""), k("c"))))
                .where(eq(column("COL2"), k("ola1")))
        );
```
results in
```
a, b, c
null
null
```
_________________________
Both cases are strange. I know that they are fixing this stuff in H2, but until the new driver is released
I think we should give H2 the same treatment as Oracle and DB2.

With this changed, and the special case for NULL addressed also in Oracle and H2, you could follow the suggestion from @joaonuno and add some test cases with NULL and empty string.
Regarding your comment there, I agree that we should not consider that mixed types are supported. For now, we should only have a guarantee for string types...

src/test/java/com/feedzai/commons/sql/abstraction/engine/impl/abs/EngineGeneralTest.java
```diff
@@ -2293,6 +2294,65 @@ public void testCaseMultipleWhenElse() throws DatabaseEngineException {
...
+
+    /**
+     * Runs a concat query on the test dataset, given a delimiter.
+     *
+     * @param delimiter the delimiter used in concat.
+     * @return the result set.
+     * @throws DatabaseEngineException if an issue when querying arises.
+     */
+    private List<Map<String, ResultColumn>> queryConcat(final Expression delimiter) throws DatabaseEngineException {
+        test6Columns();
```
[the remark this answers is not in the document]
francisco-polaco: I totally agree with you :)

francisco-polaco:
Check now

PR #188, opened 2020-12-23
Add support for IS NULL keyword

### LINKED ISSUE #186: Support Is Null operator

francisco-polaco:
In this MR, we've added support for `IS NULL` keyword.

Closes #186


francisco-polaco:
`.isNull()` is enough

PR #194, opened 2021-01-05
Support table creation DDL translation

### LINKED ISSUE #191: Support translating table creation DDL
Support translating table creation DDL, so users can export the DDL of their tables.

francisco-polaco:
Closes #191

### DIFF
--- a/src/main/java/com/feedzai/commons/sql/abstraction/engine/AbstractTranslator.java
+++ b/src/main/java/com/feedzai/commons/sql/abstraction/engine/AbstractTranslator.java
@@ -45,7 +45,7 @@ import com.feedzai.commons.sql.abstraction.dml.With;
 import com.feedzai.commons.sql.abstraction.dml.dialect.SqlBuilder;
 import com.feedzai.commons.sql.abstraction.engine.configuration.PdbProperties;
 import com.google.common.base.Joiner;
-import com.google.common.collect.ImmutableList;
+import java.util.Collections;
 import com.google.inject.Inject;
 import com.google.inject.Injector;
 import org.apache.commons.lang3.StringUtils;
@@ -685,6 +685,6 @@ public abstract class AbstractTranslator {
      */
     public List<String> translateCreateSequences(DbEntity entity) {
         // the majority of engines don't need additional SQL to create sequences.
-        return ImmutableList.of();
+        return Collections.emptyList();
     }
 }
--- a/src/main/java/com/feedzai/commons/sql/abstraction/engine/AbstractTranslator.java
+++ b/src/main/java/com/feedzai/commons/sql/abstraction/engine/AbstractTranslator.java
@@ -639,4 +639,5 @@ public abstract class AbstractTranslator {
         return "";
     }
     public abstract String translatePrimaryKeysConstraints(DbEntity entity);
+    public abstract List<String> translateForeignKey(DbEntity entity);
 }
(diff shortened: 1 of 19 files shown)


francisco-polaco:
Closes #191

src/main/java/com/feedzai/commons/sql/abstraction/engine/AbstractTranslator.java
```diff
@@ -631,4 +633,58 @@ protected Union rowsToUnion(final List<Expression> rows) {
...
+
+    /**
+     * Translates the sequence creation of the given entity table to the current dialect.
+     *
+     * @param entity The entity to translate.
+     * @return The sequence creation translation result.
+     */
+    public List<String> translateCreateSequences(DbEntity entity) {
+        // the majority of engines don't need additional SQL to create sequences.
+        return ImmutableList.of();
```
jcsf, line 688: Use 
```suggestion
        return Collections.emptyList();
```

src/main/java/com/feedzai/commons/sql/abstraction/engine/AbstractTranslator.java
```diff
@@ -44,6 +45,7 @@
 import com.feedzai.commons.sql.abstraction.dml.dialect.SqlBuilder;
 import com.feedzai.commons.sql.abstraction.engine.configuration.PdbProperties;
 import com.google.common.base.Joiner;
+import com.google.common.collect.ImmutableList;
```
jcsf, line 48:
```suggestion
import java.util.Collections;
```

jcsf (approved):

src/main/java/com/feedzai/commons/sql/abstraction/engine/impl/OracleEngine.java
```diff
@@ -457,31 +457,7 @@ protected void createTable(final DbEntity entity) throws DatabaseEngineException
...
-            }
-            columns.add(join(column, " "));
-        }
-        createTable.add("(" + join(columns, ", ") + ")");
-        // segment creation deferred can cause unexpected behaviour in sequences
-        // see: http://dirknachbar.blogspot.pt/2011/01/deferred-segment-creation-under-oracle.html
-        createTable.add("SEGMENT CREATION IMMEDIATE");
-        return join(createTable, " ");
+        return translator.translateCreateTable(entity);
     }
```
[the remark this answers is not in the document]
jmf-tls: Also, on the `createTable`method, right now we have this:
```
defaultCreateTableStatement = get createTable from translator
createTableCompressLOBS = getCompressLobsStatements(entity);
createTableStatement = defaultCreateTableStatement + createTableCompressLOBS;
```
Then we try to create the table with `createTableStatement`, which may or may not contain LOB options depending on PDB config to compress LOBs (if the option is disabled, this statement is the same as `defaultCreateTableStatement`).
If this create table fails because the DB doesn't support LOBs, then we have a retry with the default statement (without LOBs).

I would propose a change to this logic: have a single statement to create tables, with or without LOB options according to the config. This statement would be produced at `translator.translateCreateTable(entity)` as it is now, but would include the LOB options (if applicable).
This would make the logic much simpler, the method `getCompressLobsStatements` would go away and its code integrated in the `translateCreateTable`.

As it is now, with "compress LOBs" enabled PDB will either create the table with LOB compression, or if the DB doesn't support it, will create the table without compression.... with the proposed change, in the same scenario PDB would will either create the table with LOB compression or fail.
However, I believe that this is acceptable: it is up to whoever sets up the environment to make sure that the DB supports LOB compression, or if it doesn't, configure PDB with LOB compression disabled.
What do you think?
Also, @jcsf @joaonuno

jmf-tls (changes requested):
Some things are not working properly and need to be changed...

PR #196, opened 2021-01-11
Prevent dropping and adding all foreign keys when updating an entity.

upsidedownsmile:
All the foreign keys were being dropped and added unnecessarily when an entity would be updated. This would result in running more DDL statements then needed.

To prevent this from happening a diff will be computed between the current entity foreign keys and the new one and only add or remove the differences.

### DIFF
--- a/src/main/java/com/feedzai/commons/sql/abstraction/ddl/DbFk.java
+++ b/src/main/java/com/feedzai/commons/sql/abstraction/ddl/DbFk.java
@@ -15,14 +15,13 @@
  */
 package com.feedzai.commons.sql.abstraction.ddl;
 
-import com.google.common.base.Objects;
+import com.google.common.base.Preconditions;
 import com.google.common.collect.ImmutableList;
 
 import java.io.Serializable;
-import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collection;
-import java.util.Collections;
+import java.util.LinkedList;
 import java.util.List;
 import java.util.Objects;
 
@@ -52,7 +51,7 @@ public class DbFk implements Serializable {
     /**
      * Cached hashcode.
      */
-    private int hashCode = -1;
+    private final int hashCode;
 
     /**
      * Creates a new instance of {@link DbFk}.
@@ -63,6 +62,7 @@ public class DbFk implements Serializable {
         this.localColumns = ImmutableList.copyOf(builder.localColumns);
         this.referencedColumns = ImmutableList.copyOf(builder.referencedColumns);
         this.referencedTable = builder.referencedTable;
+        this.hashCode = Objects.hash(this.localColumns, this.referencedColumns, this.referencedTable);
     }
 
     /**
@@ -127,26 +127,23 @@ public class DbFk implements Serializable {
     }
 
     @Override
-    public boolean equals(final Object o) {
-        if (this == o) {
+    public boolean equals(final Object obj) {
+        if (this == obj) {
             return true;
         }
-        if (o == null || getClass() != o.getClass()) {
+
+        if (obj == null || getClass() != obj.getClass()) {
             return false;
         }
-        final DbFk dbFk = (DbFk) o;
-        return Objects.equal(getLocalColumns(), dbFk.getLocalColumns())
-                && Objects.equal(getForeignColumns(), dbFk.getForeignColumns())
-                && Objects.equal(getForeignTable(), dbFk.getForeignTable());
+
+        final DbFk dbFk = (DbFk) obj;
+        return this.localColumns.equals(dbFk.localColumns)
+                && this.referencedColumns.equals(dbFk.referencedColumns)
+                && this.referencedTable.equals(dbFk.referencedTable);
     }
 
     @Override
     public int hashCode() {
-        // Computing the hashcode might be an expensive operation because it goes through 2 lists,
-        // that's why it is being cached.
-        if(this.hashCode == -1) {
-            this.hashCode = Objects.hashCode(getLocalColumns(), getForeignColumns(), getForeignTable());
-        }
         return this.hashCode;
     }
 
@@ -154,8 +151,8 @@ public class DbFk implements Serializable {
      * Builder to create immutable {@link DbFk} objects.
      */
     public static class Builder implements com.feedzai.commons.sql.abstraction.util.Builder<DbFk>, Serializable {
-        private final List<String> localColumns = new ArrayList<>();
-        private final List<String> referencedColumns = new ArrayList<>();
+        private final List<String> localColumns = new LinkedList<>();
+        private final List<String> referencedColumns = new LinkedList<>();
         private String referencedTable = null;
 
         /**
@@ -261,7 +258,11 @@ public class DbFk implements Serializable {
 
         @Override
         public DbFk build() {
-            Objects.requireNonNull(referencedTable, "The referenced table can't be null");
+            Preconditions.checkNotNull(referencedTable, "The referenced table can't be null.");
+            Preconditions.checkArgument(localColumns.size() == referencedColumns.size(),
+                    "The number of local columns and referenced columns must be the same. local: %s ; referenced: %s",
+                    localColumns, referencedColumns
+            );
             return new DbFk(this);
         }
     }
--- a/src/main/java/com/feedzai/commons/sql/abstraction/ddl/DbFk.java
+++ b/src/main/java/com/feedzai/commons/sql/abstraction/ddl/DbFk.java
@@ -15,12 +15,14 @@
  */
 package com.feedzai.commons.sql.abstraction.ddl;
 
+import com.google.common.base.Objects;
 import com.google.common.collect.ImmutableList;
 
 import java.io.Serializable;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collection;
+import java.util.Collections;
 import java.util.List;
 import java.util.Objects;
 
@@ -46,6 +48,11 @@ public class DbFk implements Serializable {
      * The referenced table.
      */
     private final String referencedTable;
+
+    /**
+     * Cached hashcode.
+     */
+    private int hashCode = -1;
 
     /**
      * Creates a new instance of {@link DbFk}.
@@ -117,6 +124,30 @@ public class DbFk implements Serializable {
      */
     public String getReferencedTable() {
         return referencedTable;
+    }
+
+    @Override
+    public boolean equals(final Object o) {
+        if (this == o) {
+            return true;
+        }
+        if (o == null || getClass() != o.getClass()) {
+            return false;
+        }
+        final DbFk dbFk = (DbFk) o;
+        return Objects.equal(getLocalColumns(), dbFk.getLocalColumns())
+                && Objects.equal(getForeignColumns(), dbFk.getForeignColumns())
+                && Objects.equal(getForeignTable(), dbFk.getForeignTable());
+    }
+
+    @Override
+    public int hashCode() {
+        // Computing the hashcode might be an expensive operation because it goes through 2 lists,
+        // that's why it is being cached.
+        if(this.hashCode == -1) {
+            this.hashCode = Objects.hashCode(getLocalColumns(), getForeignColumns(), getForeignTable());
+        }
+        return this.hashCode;
     }
 
     /**
--- a/src/main/java/com/feedzai/commons/sql/abstraction/engine/AbstractDatabaseEngine.java
+++ b/src/main/java/com/feedzai/commons/sql/abstraction/engine/AbstractDatabaseEngine.java
@@ -42,6 +42,7 @@ import com.feedzai.commons.sql.abstraction.util.PreparedStatementCapsule;
 import com.google.common.collect.ImmutableList;
 import com.google.inject.Inject;
 import com.google.inject.Injector;
+import org.apache.commons.lang3.NotImplementedException;
 import org.apache.commons.lang3.StringUtils;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
@@ -66,6 +67,7 @@ import java.util.ArrayList;
 import java.util.HashMap;
 import java.util.HashSet;
 import java.util.LinkedHashMap;
+import java.util.LinkedList;
 import java.util.List;
 import java.util.Map;
 import java.util.Optional;
@@ -74,6 +76,7 @@ import java.util.Set;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Executors;
 import java.util.concurrent.TimeUnit;
+import java.util.stream.Collectors;
 
 import static com.feedzai.commons.sql.abstraction.engine.configuration.PdbProperties.ENCRYPTED_PASSWORD;
 import static com.feedzai.commons.sql.abstraction.engine.configuration.PdbProperties.ENCRYPTED_USERNAME;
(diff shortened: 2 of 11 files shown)


1
```diff
@@ -0,0 +1 @@
+test.txt
```
nuno-v-santos, line 1: What is this test.txt ?

src/main/java/com/feedzai/commons/sql/abstraction/ddl/DbFk.java
```diff
@@ -52,8 +59,11 @@
      * @param foreignTable   The foreign table.
      */
     private DbFk(List<String> localColumns, List<String> foreignColumns, String foreignTable) {
-        this.localColumns = localColumns;
-        this.foreignColumns = foreignColumns;
+        // Sorting to make sure that equal method run in O(n) instead of O(n^2) for comparing 2 lists.
+        Collections.sort(localColumns);
+        Collections.sort(foreignColumns);
+        this.localColumns = ImmutableList.copyOf(localColumns);
+        this.foreignColumns = ImmutableList.copyOf(foreignColumns);
```
nuno-v-santos, line 66: Should we clone first and then sort ? To avoid changing the input lists.

src/main/java/com/feedzai/commons/sql/abstraction/engine/AbstractDatabaseEngine.java
```diff
@@ -1382,12 +1461,24 @@ public synchronized ResultIterator iterator(String query, int fetchSize, int rea
...
      * @param entity The entity.
      * @throws DatabaseEngineException If something goes wrong creating the FKs.
      */
-    protected abstract void addFks(final DbEntity entity) throws DatabaseEngineException;
+    protected void addFks(final DbEntity entity) throws DatabaseEngineException {
+        addFks(entity, new HashSet<>(entity.getFks()));
+    }
+
+    /**
+     * Adds foreign keys the foreign keys passed as argument to the entity.
```
nuno-v-santos, line 1474: Review this sentence

src/main/java/com/feedzai/commons/sql/abstraction/engine/AbstractDatabaseEngine.java
```diff
@@ -1382,12 +1461,24 @@ public synchronized ResultIterator iterator(String query, int fetchSize, int rea
...
+     *
+     * @param entity The entity             .
+     * @param fks    The foreign keys to be added.
+     *
+     * @throws DatabaseEngineException If something goes wrong creating the FKs.
+     */
+    protected abstract void addFks(final DbEntity entity, final Set<DbFk> fks) throws DatabaseEngineException;
```
nuno-v-santos, line 1481: Is this a breaking change?
If `AbstractDatabaseEngine` is public api, then adding this method will break implementations of it.
Should we add a default implementation here (even if it does nothing, or throws a `NotImplementedException`) ?

1
upsidedownsmile: It should have been removed. It was a file that I was using for tests.

src/main/java/com/feedzai/commons/sql/abstraction/ddl/DbFk.java
upsidedownsmile: I don't think that is necessary in this case. The order of the items in those lists is irrelevant.

jmf-tls:
You're using a code style different from what is being used in PDB, which will cause the imports to be reorganized unnecessarily for every file you touch (and thus lots of changes in the review).

For this project, please change the order of imports in the code style to:
_import all other_
\<blank line>
_import javax.*_
_import java.*_
\<blank line>
_import static all other_

nuno-v-santos:
Do you know why there's some tests failing? Are these flaky?

upsidedownsmile:
> You're using a code style different from what is being used in PDB, which will cause the imports to be reorganized unnecessarily for every file you touch (and thus lots of changes in the review).
> 
> For this project, please change the order of imports in the code style to:
> _import all other_
> <blank line>
> _import javax.*_
> _import java.*_
> <blank line>
> _import static all other_

It should be fixed now.

nuno-v-santos (approved):

upsidedownsmile:
> Do you know why there's some tests failing? Are these flaky?

Yes it is flaky. @jmf-tls already created a PR with the fix. https://github.com/feedzai/pdb/pull/197

src/main/java/com/feedzai/commons/sql/abstraction/ddl/DbFk.java
jmf-tls: You can't sort these columns, at least like this.
It would be ok if this class was used only for comparison, but it also contains information to build foreign key constraints when creating a table.
Each local column should be considered a pair along with a "foreign column"
(in SQL a foreign column is actually a local column, so this designation is actually wrong ─ see #201 )

For example, to get this
`FOREIGN KEY (b, c) REFERENCES other_table (c1, c2)`
we would add `b` and `c` to `localColumns`
and `c1` and `c2` to `foreignColumns`.
Columns `b` and `c1` need to be paired: the data types in these columns have to match, and values inserted in `b` need to be present in `c1`.
Same for the pair `c` and `c2` (they may have the same data type or values as `c1` for example, but that would be just a coincidence, not a requirement).

Even keeping the order `b,c` and `c1,c2`, instead of having `c,b` and `c2,c1`, may be important, but I'm not sure about this.

src/main/java/com/feedzai/commons/sql/abstraction/ddl/DbFk.java
```diff
@@ -150,7 +184,7 @@ public Builder addForeignColumns(final Collection<String> columns) {
 
         @Override
         public DbFk build() {
-            return new DbFk(ImmutableList.copyOf(localColumns), ImmutableList.copyOf(foreignColumns), foreignTable);
+            return new DbFk(localColumns, foreignColumns, foreignTable);
```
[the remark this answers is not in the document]
jmf-tls: I already did this on #201 
You won't need to do it yourself, but you'll still need to rebase :smiley_cat:

src/main/java/com/feedzai/commons/sql/abstraction/engine/AbstractDatabaseEngine.java
```diff
@@ -654,6 +632,108 @@ public synchronized void updateEntity(DbEntity entity) throws DatabaseEngineExce
...
+        final Map<String, DbFk.Builder> existentFks = new HashMap<>();
+
+        try {
+            getConnection();
+            ResultSet rs = getMetadata().getImportedKeys(null, this.currentSchema, entityName);
+            while (rs.next()) {
+                final String fkName = rs.getString("FK_NAME");
+                final String foreignTable = rs.getString("FKTABLE_NAME");
+                final String localColumnName = rs.getString("PKCOLUMN_NAME");
+                final String foreignColumnName = rs.getString("FKCOLUMN_NAME");
```
jmf-tls, line 717: I have tested this and got wrong results.
You probably got confused due to wrong names adopted in the class `DbFk`, and because of that I created #201 to fix the issue.

In this statement
```
ALTER TABLE mytable ADD CONSTRAINT pk_123 FOREIGN KEY (b, c) REFERENCES other_table (c1, c2)
```
the SQL standard calls "mytable" the foreign table, and the foreign key is composed by the columns "b" and "c"
(as it is clear in the statement).

So, while here we would expect to get
- foreignTable = other_table
- localColumnName = b
- foreignColumnName = c1

we will actually get
- foreignTable = mytable
- localColumnName = c1
- foreignColumnName = b

After #201 is landed, you can rebase, and then use
```suggestion
                final String fkName = rs.getString("FK_NAME");
                final String localColumnName = rs.getString("FKCOLUMN_NAME");
                final String referencedTable = rs.getString("PKTABLE_NAME");
                final String referencedColumnName = rs.getString("PKCOLUMN_NAME");
```

src/main/java/com/feedzai/commons/sql/abstraction/engine/AbstractDatabaseEngine.java
```diff
@@ -654,6 +632,108 @@ public synchronized void updateEntity(DbEntity entity) throws DatabaseEngineExce
...
+                        .foreignTable(foreignTable)
+                        .addColumn(localColumnName)
+                        .addForeignColumn(foreignColumnName)
+                        : fk.foreignTable(foreignTable)
+                            .addColumn(localColumnName)
+                            .addForeignColumn(foreignColumnName)
+                );
+            }
+
+            return existentFks.entrySet().stream().collect(Collectors.toMap(Map.Entry::getKey, entry -> entry.getValue().build()));
```
jmf-tls, line 730: long line, please break at least the `collect` into a new line

src/main/java/com/feedzai/commons/sql/abstraction/engine/AbstractDatabaseEngine.java
```diff
@@ -654,6 +632,108 @@ public synchronized void updateEntity(DbEntity entity) throws DatabaseEngineExce
...
+
+                existentFks.compute(fkName, (name, fk) -> fk == null
+                        ? new DbFk.Builder()
+                        .foreignTable(foreignTable)
+                        .addColumn(localColumnName)
+                        .addForeignColumn(foreignColumnName)
+                        : fk.foreignTable(foreignTable)
+                            .addColumn(localColumnName)
+                            .addForeignColumn(foreignColumnName)
+                );
```
jmf-tls, line 727: confusing...
Considering that a given `fkName` can only have 1 "foreignTable",
and assuming that  #201 is landed, you have rebased, and the proposed changes from the comment above have been accepted,
how about this?
```suggestion
                existentFks.computeIfAbsent(fkName, name -> new DbFk.Builder().referencedTable(referencedTable))
                        .addColumn(localColumnName)
                        .addReferencedColumn(referencedColumnName);
```

jmf-tls (changes requested):
Some things are not working properly, mostly due to a mismatch between the names used in PDB and those adopted in SQL standards.
In particular, the meaning of foreign key/column/table.
This problem is addressed in #201 , please check.
Let's get that one landed and then rebase here.

src/main/java/com/feedzai/commons/sql/abstraction/engine/AbstractDatabaseEngine.java
jmf-tls: I think we're considering only the `DatabaseEngine` as part of the public api, not the abstract class or the other engine implementation classes.

jmf-tls (approved):

nuno-v-santos (approved):

src/test/java/com/feedzai/commons/sql/abstraction/engine/impl/abs/ForeignKeyTest.java
```diff
@@ -0,0 +1,474 @@
...
+    /**
+     * Helper method to verify the behavior of an entity with a foreign key.
+     * <p>
+     * This method performs the following actions:
+     * <ol>
+     *     <li>adds the entity with a foreign key constraint to the database (the primary entities are expected
+     *     to be there already)</li>
+     *     <li>checks that the foreign key constraint prevents the "invalid entry" from being persisted</li>
+     *     <li>persists the "valid entry"</li>
+     *     <li>checks that the foreign key constraint prevents rows referenced by the "valid entry" from being deleted
```
upsidedownsmile, line 243: Should this be tested here? This depends on the policy used for `on delete` and not the foreign key. Since we don't explicitly set the policy, different DBs might have different defaults and these tests might fail in the future if we add support for other DBs.

This also applies to `removeRowPreviouslyReferencedByForeignKeyTest` and `removeRowAfterRemovingForeignKeyTest`

src/main/java/com/feedzai/commons/sql/abstraction/engine/impl/DB2Engine.java
```diff
@@ -67,32 +70,32 @@
...
      */
-    public static final String FOREIGN_ALREADY_EXISTS = "DB2 SQL Error: SQLCODE=-601, SQLSTATE=42710";
+    static final String DB2_DEFAULT_BLOB_SIZE = "2G";
 
     /**
-     * The default size of a BLOB in DB2.
+     * The column must have a {@link DbColumnConstraint#NOT_NULL NOT NULL column constraint}.
      */
-    public static final String DB2_DEFAULT_BLOB_SIZE = "2G";
+    private static final int COLUMN_MUST_BE_NON_NULL = -542;
```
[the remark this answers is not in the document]
jmf-tls: I'd say that what makes the public API in PDB is everything except DB-specific code:
interfaces, enums, exceptions, and all dml and ddl classes that are not some DB implementation.  
Maybe one day we can separate this public stuff from the rest into separate modules, and do breaking changes checks :smile: 

This is an implementation for DB2, and not only we should rely on the SQL code instead of error messages too check what type of error is occurring on the DB, but also there's no reason for these codes to be public.

src/test/java/com/feedzai/commons/sql/abstraction/engine/impl/abs/ForeignKeyTest.java
jmf-tls: I think it's a good way to test:
- a foreign key constraint should always prevents the "invalid entry" (where the entry references a value that doesn't exist in the reference table) from being persisted, regardless of the ON DELETE/UPDATE setting;
- as strange as it may seem :smile: the DB vendors actually agreed on this one (I don't know if it's explicitly stated on the SQL standard), by default we don't have ON DELETE CASCADE, so deleting a referenced row that is referenced by a foreign key results in an error for all of them; in the future we may add support for that, in which case the tests will need to be updated, they will probably fail

src/main/java/com/feedzai/commons/sql/abstraction/engine/impl/OracleEngine.java
```diff
@@ -1033,6 +1039,55 @@ protected void addFks(DbEntity entity) throws DatabaseEngineException {
...
+                "  AND fc.owner = f.owner\n" +
+                "  AND fc.constraint_name = f.constraint_name\n" +
+                "  AND fc.table_name = f.table_name\n" +
+                "  AND fc.position = pc.position\n" +
+                "  ORDER BY pktable_schem, pktable_name, key_seq";
+
+        final PreparedStatement ps = this.conn.prepareStatement(query);
+        ps.setString(1, entityName);
+        ps.setString(2, this.currentSchema);
+        ps.closeOnCompletion();
```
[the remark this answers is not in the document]
jmf-tls: This is what Oracle driver is doing, and I believe it is the correct thing to do.
This method is supposed to replace the `java.sql.DatabaseMetaData#getImportedKeys` that all other engines are using, and thus returns a ResultSet.
The statement `ps.closeOnCompletion()` makes sure that the PreparedStatement is closed when the ResultSet is closed.
All usages of this `getImportedKeys` method use the try-with-resources on the ResultSet, and should close the statement; if we did try-with-resources on the PreparedStatement here, some values might be no longer available on the ResultSet.

src/main/java/com/feedzai/commons/sql/abstraction/engine/impl/DB2Engine.java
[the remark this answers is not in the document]
upsidedownsmile: > This is an implementation for DB2, and not only we should rely on the SQL code instead of error messages too check what type of error is occurring on the DB, but also there's no reason for these codes to be public.

Completly agree, this way is so much better than before, I was just worried that we would break something.

PR #197, opened 2021-01-12
Fix BatchUpdateTest flakyness

jmf-tls:
Summary:
This test has been failing many times, even when there are no code changes.

Looks like the timed pauses don't offer enough guarantees on the test environment;
this commit removes the dependency on timings by using latches.

Also, we don't have guarantees that the first thread, as soon as it releases the
`flushTransactionLock` will immediately add "first" to the results list, before
the second thread does it (only the first thread really does a flush on the DB).
For that reason, only the flushes that don't have anything in the batch and thus
don't do a flush on the DB are compared (because "second" uses a synchronized
flush, it should always finish after the "third").

### DIFF
--- a/src/main/java/com/feedzai/commons/sql/abstraction/batch/AbstractBatch.java
+++ b/src/main/java/com/feedzai/commons/sql/abstraction/batch/AbstractBatch.java
@@ -403,16 +403,23 @@ public abstract class AbstractBatch implements Runnable {
     /**
      * Flushes the pending messages.
      * <p>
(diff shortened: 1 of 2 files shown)


TiagoMSSantos (approved):
Looks good to me.

hugomiguelabreu (approved):

jmf-tls:
Ran the tests 3 times, now all of them have always passed.

TiagoMSSantos (approved):

hugomiguelabreu (approved):

PR #201, opened 2021-01-19
Add proper names to foreign key related information

jmf-tls:
Summary:
According to SQL documentation, for example in MySQL
"A foreign key relationship involves a parent table that holds the initial
column values, and a child table with column values that reference the parent
column values. A foreign key constraint is defined on the child table."

The method `Connection.getMetadata().getImportedKeys(...)` returns foreign key
metadata that conforms to that definition:
- the columns with name starting with FKTABLE_ will contain information about
 the foreign key table, i.e. the child table (the table where the foreign key
 relation is defined);
- the columns with name starting with PKTABLE_ will contain information about
 the primary key table, i.e. the parent table (the table whose columns are
 referenced by the columns in the child table).

In PDB the object that represents a foreign key is DbFk.
Previously it contained the fields "foreignColumns" and "foreignTable", to
refer to the parent table. In fact, according to the SQL standard, foreign
columns are the columns that compose the foreign key, and their name implies
that they refer to local columns.
Not only this is wrong, but it is also confusing and error prone, as it may
lead to bugs when relying on these names and then calling JDBC methods where
the meaning of those names is different.

This commit deprecates those fields and respective getter and builder methods,
and replaces them with unambiguous "referenced columns/tables" designations.


jmf-tls:
This is related to #196
Due to the names used in DbFk, a new method introduced in that review that was using JDBC's `getImportedKeys` method was actually swapping the results: the column names that corresponded to the local table were being set as if they were from the referenced table, and vice-versa...

joaonuno (approved):

nuno-v-santos (approved):

PR #210, opened 2021-02-01
Bump JMockit to 1.49

jmf-tls:
Summary:
Bumps JMockit version in preparation for Java 11 and beyond.

### DIFF
--- a/src/test/java/com/feedzai/commons/sql/abstraction/engine/impl/abs/AbstractEngineSchemaTest.java
+++ b/src/test/java/com/feedzai/commons/sql/abstraction/engine/impl/abs/AbstractEngineSchemaTest.java
@@ -19,6 +19,7 @@ import com.feedzai.commons.sql.abstraction.ddl.DbColumnType;
 import com.feedzai.commons.sql.abstraction.ddl.DbEntity;
 import com.feedzai.commons.sql.abstraction.ddl.DbEntityType;
 import com.feedzai.commons.sql.abstraction.dml.result.ResultColumn;
+import com.feedzai.commons.sql.abstraction.engine.AbstractDatabaseEngine;
 import com.feedzai.commons.sql.abstraction.engine.ConnectionResetException;
 import com.feedzai.commons.sql.abstraction.engine.DatabaseEngine;
 import com.feedzai.commons.sql.abstraction.engine.DatabaseEngineException;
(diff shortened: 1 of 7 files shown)


src/test/java/com/feedzai/commons/sql/abstraction/engine/impl/abs/AbstractEngineSchemaTest.java
```diff
@@ -683,8 +683,9 @@ private static void checkEntity(final String reasonMessage, final DbEntity entit
                 // this second entry is needed because MySQL returns tables as SYSTEM_TABLEs (when they are created in default schema)
                 MapEntry.entry(name, DbEntityType.SYSTEM_TABLE)
         };
-        final String originalSchema = Deencapsulation.getField(originalEngine, "currentSchema");
-        final String otherSchema = Deencapsulation.getField(otherEngine, "currentSchema");
```
jmf-tls, line 687: These don't exist anymore on the newest version.
We would need to get the `Field` and then get the value ourselves....

It's just easier to use the already existing `getSechema()`

src/test/java/com/feedzai/commons/sql/abstraction/engine/impl/abs/EngineGeneralTest.java
```diff
@@ -4121,34 +4122,35 @@ public void entityEntryHashcodeTest() {
      * exited from.
      *
      * @since 2.1.12
+     * @throws Exception if something goes wrong while checking if the connection of the engine is closed.
      */
     @Test
-    public void tryWithResourcesClosesEngine() {
```
jmf-tls, line 4126: This test was relying on `MockUp.getMockInstance()`, which doesn't exist anymore on the newest version.

Initially I removed it, since it was just verifying Java functionality and not PDB, and was equivalent to simply doing
`assertTrue(AutoCloseable.class.isAssignableFrom(DatabaseEngine.class))`

I restored the test because we can test that the DatabaseEngine behaves as expected in a try-with-resources,
by checking that it closes the underlying DB connection.

src/test/java/com/feedzai/commons/sql/abstraction/engine/impl/abs/AbstractEngineSchemaTest.java
hugomiguelabreu: Yeah, not sure why wasn't public in the first place. Do you know of any concern?

hugomiguelabreu (approved):

src/test/java/com/feedzai/commons/sql/abstraction/engine/impl/abs/AbstractEngineSchemaTest.java
jmf-tls: I thought it was me, since I did a fix for PDB to properly use the schema some time ago,
but that `getSchema` method exists since this repository was created in 2014!

I don't see any problem being public, it's only "public" in the engine classes but not in the interface.
It has no use in the interface anyway, because we already know it beforehand, as we specify it in the config.

TiagoMSSantos (approved):

PR #212, opened 2021-02-05
Add support for CAST to MySQL

jmf-tls:
Summary:
MySQL was left out when CAST was introduced.
This is not a desired situation, because PDB aims to present
features that work across all supported DB engines seamlessly.
Since MySQL does in fact support CAST, this commit adds it to
that engine.
The tests for CAST were also improved.

### DIFF
--- a/src/main/java/com/feedzai/commons/sql/abstraction/engine/impl/MySqlTranslator.java
+++ b/src/main/java/com/feedzai/commons/sql/abstraction/engine/impl/MySqlTranslator.java
@@ -282,7 +282,30 @@ public class MySqlTranslator extends AbstractTranslator {
 
     @Override
     public String translate(final Cast cast) {
-        throw new OperationNotSupportedRuntimeException("PDB does not support Cast in MySQL.");
+        final String type;
+
+        switch (cast.getType()) {
+            case BOOLEAN:
+            case INT:
+            case LONG:
+                type = "SIGNED INTEGER";
+                break;
+            case DOUBLE:
+                type = "DECIMAL(65,30)";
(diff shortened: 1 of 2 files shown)


src/main/java/com/feedzai/commons/sql/abstraction/engine/impl/MySqlTranslator.java
```diff
@@ -282,7 +282,30 @@ public String translate(DbColumn c) {
...
+            case BOOLEAN:
+            case INT:
+            case LONG:
+                type = "SIGNED INTEGER";
+                break;
+            case DOUBLE:
+                type = "DECIMAL(65,30)";
+                break;
+            case STRING:
+                type = "CHAR";
```
hugomiguelabreu, line 297: Not VARCHAR?

jmf-tls: as strange as it may seem, in MySQL 5.7 (and v8.0 doesn't add much to this) we can only cast to:
BINARY
CHAR
DATE
DATETIME
DECIMAL
JSON
NCHAR
SIGNED [INTEGER]
TIME
UNSIGNED [INTEGER]

jmf-tls: (but it's like casting to a string, it gets converted with whatever characters it needs, doesn't get truncated)

upsidedownsmile (approved):

hugomiguelabreu (approved):

PR #231, opened 2021-04-08
PDB-24: Implement connection pool

### LINKED ISSUE #24: Support DatabaseEngine pools.
I always find myself adding custom code to add support for connection pooling.
Native PDB support for pooling would be a time saver.

francisco-polaco:
Hello guys!

I'm accepting test suggestions. 😄 
Keep in mind that we are already using a well tested ObjectPool (Apache Commons Pool) and PDB itself. So the tests should be more focused in the integration between these 2 components.

As a side bonus, this new connection pool was successfully tested with two Feedzai products.

[Internal Issue](https://issues.feedzai.com/browse/JANUS-346)

This closes #24

### DIFF
--- a/README.md
+++ b/README.md
@@ -186,9 +186,8 @@ and `pool.generic.maxWaitMillis` are the most common ones.
 | `pool.abandoned.removeAbandonedTimeout` | Timeout in seconds before an abandoned `DatabaseEngine` instance can be removed. | 300 |
 | `pool.abandoned.logAbandoned` | Flag to log stack traces for application code which abandoned a `DatabaseEngine` instance. | false |
 
-{% hint style='info' %}
-If no `pool.abandoned` property is defined in the configuration file, then no policy to remove abandoned `DatabaseEngine` instances is applied.
-{% endhint %}
+> If no `pool.abandoned` property is defined in the configuration file, then no policy to remove abandoned `DatabaseEngine` instances is applied.
+
 ### Create Table
 
 We start by creating the table to store the different data Types:
--- a/README.md
+++ b/README.md
@@ -145,6 +145,50 @@ It is also important to select a schema policy. There are four possible schema p
 - drop-create - New entities are dropped before creation if they already exist.
 - none - The program is not allowed to create new entities.
 
+### PDB Pool Usage
+PDB natively supports connection pools. You can create one using either a `Properties` or a `Map<String, String>`
+instance when defining the properties.
+```java
+final DatabaseEnginePool dbPool = DatabaseEnginePool.getConnectionPool(properties);
+```
+
+Additionally, you can specify a modifier for the `DatabaseEngine`.
+This can be useful when you need to load entities to be able to insert entries on them.
+
+```java
+final DatabaseEnginePool dbPool = DatabaseEnginePool.getConnectionPool(properties, engine -> engine.loadEntity(entity));
+```
+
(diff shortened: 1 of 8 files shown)


francisco-polaco:
Hello guys!

I'm accepting test suggestions. Keep in mind that we are already using a well tested ObjectPool (Apache Commons Pool) and PDB itself. So the tests should be more focused in the integration between these 2 components.

As a side bonus, this new connection pool was successfully tested with a Feedzai product.

README.md
```diff
@@ -145,6 +145,50 @@ It is also important to select a schema policy. There are four possible schema p
...
+| `pool.abandoned.removeAbandonedTimeout` | Timeout in seconds before an abandoned `DatabaseEngine` instance can be removed. | 300 |
+| `pool.abandoned.logAbandoned` | Flag to log stack traces for application code which abandoned a `DatabaseEngine` instance. | false |
+
+{% hint style='info' %}
+If no `pool.abandoned` property is defined in the configuration file, then no policy to remove abandoned `DatabaseEngine` instances is applied.
+{% endhint %}
```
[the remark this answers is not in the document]
francisco-polaco: Fixed

src/main/java/com/feedzai/commons/sql/abstraction/engine/pool/PooledDatabaseEngine.java
```diff
@@ -0,0 +1,436 @@
...
+     * @param pool the pool owning the pooled database engine.
+     * @param engine the database engine to be wrapped.
+     */
+    PooledDatabaseEngine(final GenericObjectPool<PooledDatabaseEngine> pool, final DatabaseEngine engine) {
+        this.pool = pool;
+        this.engine = engine;
+    }
+
+    @Override
+    public void close() {
```
nahian18, line 86: Since closing the `PooledDatabaseEngine` does not actually close the `DatabaseEngine` object, I would like to understand/see how the database connection actually gets freed when the pooled object containing the `DatabaseEngine` object is evicted or destroyed from the pool.

francisco-polaco: Hi @nahian18 

I did some quick tests and indeed we're not closing the underlying connections after returning the `DatabaseEngine` object to the pool. 😢 

Thanks for point this! Following the [apache pool documentation](https://commons.apache.org/proper/commons-pool/apidocs/org/apache/commons/pool2/PooledObjectFactory.html), I'veoverriddend the `destroyObject` method to close the underlying connection.
After this change my tests show that we aren't leaking connections anymore 😄 

Can you please review the recent pushed changes?

Thanks!

francisco-polaco: Hi again!

Just added more logging for this methods and a unit test to check if the connection is closed once the object is destroyed!

Thanks!

nahian18 (approved):

luismcosta:
I haven't look at the code, but, strategically-wise it makes sense to this type of functionality inside PDB.
Those that today and already have their own connection pools don't have to use this one, but projects starting to depend on PDB can leverage on  this new functionality.

In fact, there are some products from Feedzai,  that mention that their internal connection pool should be factored to PDB.

PR #243, opened 2021-06-22
Allow using SELECT ... FOR UPDATE

artpdr:
This commit enables the use of SELECT ... FOR UPDATE. The use will only be worth if running
inside a transaction explicitly (e.g #beginTransaction followed by #commit or #rollback).

### DIFF
--- a/src/main/java/com/feedzai/commons/sql/abstraction/dml/Query.java
+++ b/src/main/java/com/feedzai/commons/sql/abstraction/dml/Query.java
@@ -69,6 +69,14 @@ public class Query extends Expression {
     private boolean distinct = false;
 
     /**
+     * Signals if query should include FOR UPDATE. If it is set as {@literal true} the database will block reads from another
+     * transactions also using SELECT FOR UPDATE.
+     *
+     * @since 2.8.4
+     */
+    private boolean forUpdate = false;
+
+    /**
      * Creates a new instance of {@link Query}.
      */
     public Query() {
@@ -158,6 +166,17 @@ public class Query extends Expression {
      */
     public boolean isDistinct() {
         return distinct;
+    }
+
+    /**
+     * Checks if query is a FOR UPDATE. If it is set as {@literal true} the database will block reads from another
+     * transactions also using SELECT FOR UPDATE.
+     *
+     * @return {@literal true} if it is for update; {@literal false} otherwise.
+     * @since 2.8.4
+     */
+    public boolean isForUpdate() {
+        return this.forUpdate;
     }
 
     /**
@@ -368,6 +387,20 @@ public class Query extends Expression {
             limit = Integer.MAX_VALUE - offset;
         }
 
+        return this;
+    }
+
+    /**
+     * Sets whether query is FOR UPDATE or not. If it is set as {@literal true} the database will block reads from another
+     * transactions also using SELECT FOR UPDATE.
+     *
+     * @param forUpdate If it is FOR UPDATE or not.
+     * @return This expression.
+     * @implNote When using Cockroach DB, this method has no meaning.
+     * @since 2.8.4
+     */
+    public Query forUpdate(final boolean forUpdate) {
+        this.forUpdate = forUpdate;
         return this;
     }
 }
--- a/src/test/java/com/feedzai/commons/sql/abstraction/engine/impl/abs/SelectForUpdateTest.java
+++ b/src/test/java/com/feedzai/commons/sql/abstraction/engine/impl/abs/SelectForUpdateTest.java
@@ -1,0 +1,231 @@
+/*
+ * Copyright 2021 Feedzai
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.feedzai.commons.sql.abstraction.engine.impl.abs;
+
+import com.feedzai.commons.sql.abstraction.ddl.DbColumnType;
+import com.feedzai.commons.sql.abstraction.ddl.DbEntity;
+import com.feedzai.commons.sql.abstraction.dml.result.ResultColumn;
+import com.feedzai.commons.sql.abstraction.engine.DatabaseEngine;
+import com.feedzai.commons.sql.abstraction.engine.DatabaseEngineException;
+import com.feedzai.commons.sql.abstraction.engine.DatabaseEngineTimeoutException;
+import com.feedzai.commons.sql.abstraction.engine.DatabaseFactory;
+import com.feedzai.commons.sql.abstraction.engine.impl.cockroach.SkipTestCockroachDB;
+import com.feedzai.commons.sql.abstraction.engine.testconfig.DatabaseConfiguration;
+import com.feedzai.commons.sql.abstraction.engine.testconfig.DatabaseTestUtil;
+import com.sun.org.apache.xpath.internal.operations.Bool;
+import org.assertj.core.api.Assertions;
+import org.junit.After;
+import org.junit.Before;
+import org.junit.Test;
+import org.junit.experimental.categories.Category;
+import org.junit.runner.RunWith;
+import org.junit.runners.Parameterized;
+
+import java.util.Collection;
+import java.util.List;
+import java.util.Map;
+import java.util.Properties;
+import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.CountDownLatch;
+import java.util.concurrent.Future;
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.atomic.AtomicBoolean;
+
+import static com.feedzai.commons.sql.abstraction.dml.dialect.SqlBuilder.all;
+import static com.feedzai.commons.sql.abstraction.dml.dialect.SqlBuilder.column;
+import static com.feedzai.commons.sql.abstraction.dml.dialect.SqlBuilder.entry;
+import static com.feedzai.commons.sql.abstraction.dml.dialect.SqlBuilder.eq;
+import static com.feedzai.commons.sql.abstraction.dml.dialect.SqlBuilder.k;
+import static com.feedzai.commons.sql.abstraction.dml.dialect.SqlBuilder.select;
+import static com.feedzai.commons.sql.abstraction.dml.dialect.SqlBuilder.table;
+import static com.feedzai.commons.sql.abstraction.dml.dialect.SqlBuilder.update;
+import static com.feedzai.commons.sql.abstraction.engine.configuration.PdbProperties.ENGINE;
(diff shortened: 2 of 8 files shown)


src/main/java/com/feedzai/commons/sql/abstraction/dml/Query.java
```diff
@@ -68,6 +68,13 @@
      */
     private boolean distinct = false;
 
+    /**
+     * Signals if query should include FOR UPDATE.
+     *
+     * @since ```feedzai.next.release```
```
TiagoMSSantos, line 74: I thought that we didn't use these kind of tags here.

src/test/java/com/feedzai/commons/sql/abstraction/engine/impl/abs/SelectForUpdateTest.java
```diff
@@ -0,0 +1,177 @@
...
+import static com.feedzai.commons.sql.abstraction.engine.configuration.PdbProperties.JDBC;
+import static com.feedzai.commons.sql.abstraction.engine.configuration.PdbProperties.PASSWORD;
+import static com.feedzai.commons.sql.abstraction.engine.configuration.PdbProperties.SCHEMA_POLICY;
+import static com.feedzai.commons.sql.abstraction.engine.configuration.PdbProperties.USERNAME;
+
+/**
+ * Tests the SELECT ... FOR UPDATE case.
+ *
+ * @author Artur Pedroso (artur.pedroso@feedzai.com)
+ * @since @@@feedzai.next.release@@@
```
TiagoMSSantos, line 55: I think you have to put the next version here instead of these tags.

src/test/java/com/feedzai/commons/sql/abstraction/engine/impl/abs/SelectForUpdateTest.java
```diff
@@ -0,0 +1,177 @@
...
+ */
+@RunWith(Parameterized.class)
+public class SelectForUpdateTest {
+    private static final String TEST_TABLE = "pretty_locks_table";
```
TiagoMSSantos, line 59: I would add Javadoc to these fields so its easier to understand what they mean.

src/test/java/com/feedzai/commons/sql/abstraction/engine/impl/abs/SelectForUpdateTest.java
```diff
@@ -0,0 +1,177 @@
...
+            {
+                setProperty(JDBC, config.jdbc);
+                setProperty(USERNAME, config.username);
+                setProperty(PASSWORD, config.password);
+                setProperty(ENGINE, config.engine);
+                setProperty(SCHEMA_POLICY, "create");
+            }
+        };
+
+        engine = DatabaseFactory.getConnection(properties);
```
TiagoMSSantos, line 91: I would also add the `this` keyword so its easier to read / understand that this variable is a field when not using an IDE.

src/test/java/com/feedzai/commons/sql/abstraction/engine/impl/abs/SelectForUpdateTest.java
```diff
@@ -0,0 +1,177 @@
...
+                engine2.commit();
+            } catch (final DatabaseEngineTimeoutException e) {
+                try {
+                    // As h2 has a lockTimeout = 2 seconds, we need to retry the Assert
+                    Thread.sleep(10000);
+                    Assertions.assertThat(queryForUpdate(engine2).get(0).get(COOL_COL).toString())
+                            .as("Second client should read value updated from first client after 1st client commit.")
+                            .isEqualTo(FINAL_COOL_COL_VAL);
+                    engine2.commit();
+                } catch (Exception e2) {
```
TiagoMSSantos, line 147: Missing `final` keyword. Also, don't we want to use this exception in the new one that is thrown?

src/main/java/com/feedzai/commons/sql/abstraction/dml/Query.java
hugomiguelabreu: We don't. We need to materialize the version manually.

jcsf: +1, this needs to have the next version "hardcoded"

src/test/java/com/feedzai/commons/sql/abstraction/engine/impl/abs/SelectForUpdateTest.java
jcsf: +1, this needs to have the next version "hardcoded"

src/main/java/com/feedzai/commons/sql/abstraction/dml/Query.java
```diff
@@ -160,6 +167,15 @@ public boolean isDistinct() {
         return distinct;
     }
 
+    /**
+     * Checks if query is a FOR UPDATE.
+     *
+     * @return {@code true} if it is for update; {@code false} otherwise.
```
jcsf, line 173:
```suggestion
     * @return {@literal true} if it is for update; {@literal false} otherwise.
```

src/test/java/com/feedzai/commons/sql/abstraction/engine/impl/abs/SelectForUpdateTest.java
```diff
@@ -0,0 +1,177 @@
+/*
+ * The copyright of this file belongs to Feedzai. The file cannot be
+ * reproduced in whole or in part, stored in a retrieval system,
+ * transmitted in any form, or by any means electronic, mechanical,
+ * photocopying, or otherwise, without the prior permission of the owner.
+ *
+ * © 2021 Feedzai, Strictly Confidential
+ */
```
jcsf, line 8: Not sure that PDB uses this in this way.

src/test/java/com/feedzai/commons/sql/abstraction/engine/impl/abs/SelectForUpdateTest.java
jcsf: +1

src/test/java/com/feedzai/commons/sql/abstraction/engine/impl/abs/SelectForUpdateTest.java
```diff
@@ -0,0 +1,177 @@
...
+    public void cleanup() throws Exception {
+        engine.dropEntity(TEST_TABLE);
+        engine.close();
+    }
+
+    @Test
+    // Cockroach doesn't respect FOR UPDATE. It will execute transactions with Serializable isolation and we need
+    // to handle errors when updating the same item concurrently on the client side
+    // (see https://www.cockroachlabs.com/docs/v19.2/postgresql-compatibility#locking-and-for-update).
+    @Category(SkipTestCockroachDB.class)
```
jcsf, line 122: Why don't you annotate the full class instead of only the method?

src/test/java/com/feedzai/commons/sql/abstraction/engine/impl/abs/SelectForUpdateTest.java
```diff
@@ -0,0 +1,177 @@
...
+                secondClientSelectedForUpdate.countDown();
+                engine2.commit();
+            } catch (final DatabaseEngineTimeoutException e) {
+                try {
+                    // As h2 has a lockTimeout = 2 seconds, we need to retry the Assert
+                    Thread.sleep(10000);
+                    Assertions.assertThat(queryForUpdate(engine2).get(0).get(COOL_COL).toString())
+                            .as("Second client should read value updated from first client after 1st client commit.")
+                            .isEqualTo(FINAL_COOL_COL_VAL);
+                    engine2.commit();
```
jcsf, line 146: Why don't you just put the code above in a while loop with a timeout for the test? instead of duplicating it?

src/test/java/com/feedzai/commons/sql/abstraction/engine/impl/abs/SelectForUpdateTest.java
```diff
@@ -0,0 +1,177 @@
...
+                } catch (Exception e2) {
+                    throw new RuntimeException("Unexpected exception while running second client.", e);
+                }
+            } catch (final Exception e) {
+                throw new RuntimeException("Unexpected exception while running second client.", e);
+            }
+            secondClientFinishedWithSuccess.set(true);
+        });
+        secondClient.start();
```
jcsf, line 155: Just a sugestion, you can run this in a single thread executor and with that use completable futures instead of relying on your atomic boolean to know when this is completed. Makes it easier to read.

src/test/java/com/feedzai/commons/sql/abstraction/engine/impl/abs/SelectForUpdateTest.java
```diff
@@ -0,0 +1,177 @@
...
+
+        Assertions.assertThat(secondClientSelectedForUpdate.await(10, TimeUnit.SECONDS))
```
jcsf, line 157: We really need 10 seconds? cant it just be 5? or less? it will make the tests take > 10 seconds just for a quick validation.

src/test/java/com/feedzai/commons/sql/abstraction/engine/impl/abs/SelectForUpdateTest.java
artpdr: I like your suggestion but don't like the time the change will take

src/test/java/com/feedzai/commons/sql/abstraction/engine/impl/abs/SelectForUpdateTest.java
```diff
@@ -0,0 +1,222 @@
...
+                .as("Second client should be waiting for DB lock from first client to be released.")
+                .isFalse();
+
+        engine.executeUpdate(
+                update(table(TEST_TABLE))
+                        .set(eq(column((COOL_COL)), k(FINAL_COOL_COL_VAL)))
+                        .where(eq(column(PK_COL), k(PK_COL_VAL)))
+        );
+        engine.commit();
+        Assertions.assertThat(secondClient.get()).isTrue();
```
jcsf, line 213: Put a timeout on this get.

src/test/java/com/feedzai/commons/sql/abstraction/engine/impl/abs/SelectForUpdateTest.java
```diff
@@ -0,0 +1,222 @@
...
+    @Test(timeout = 20000)
+    public void testThatSecondClientQueryReadsValueOnlyAfterFirstClientCommit() throws Exception {
+        this.engine.beginTransaction();
+        queryForUpdate(this.engine);
+
+        final CountDownLatch secondClientSelectedForUpdate = new CountDownLatch(1);
+        final Future<Boolean> secondClient = CompletableFuture.supplyAsync(() -> {
+            try {
+                final DatabaseEngine engine2 = DatabaseFactory.getConnection(properties);
+                while (true) {
```
jcsf, line 185: Put a boolean that you can control here so that in the end of the test this is interrupted.
Otherwise this may be running forever in the jvm.

src/test/java/com/feedzai/commons/sql/abstraction/engine/impl/abs/SelectForUpdateTest.java
artpdr: We have the test timeout

TiagoMSSantos (approved):
lgtm

jcsf (approved):

PR #245, opened 2021-06-28
Add support for JDK 11 and 17

jcsf:
**Summary:**
In order to prepare the JDK upgrade, new builds are being added which compile and test this
project against JDK 11 and JDK 17.

### DIFF
--- a/.github/workflows/build-pdb.yml
+++ b/.github/workflows/build-pdb.yml
@@ -30,3 +30,51 @@ jobs:
       
     - name: Build with Maven
       run: mvn -B clean install --file pom.xml -DskipTests
+
+  build-jdk11:
+
+    runs-on: ubuntu-latest
+
+    steps:
+      - uses: actions/checkout@v2
+      - name: Set up JDK 11
+        uses: actions/setup-java@v1
+        with:
+          java-version: 11
+
+      - name: Cache Maven packages
+        uses: actions/cache@v2.1.4
+        with:
+          path: ~/.m2
+          key: ${{ runner.os }}-m2-${{ hashFiles('**/pom.xml') }}-jdk11
+          restore-keys: ${{ runner.os }}-m2-jdk11
+
+      - name: Build with Maven
+        run: mvn -B clean install --file pom.xml -DskipTests
+
+  build-jdk17:
+
+    runs-on: ubuntu-latest
+
+    steps:
+      - uses: actions/checkout@v2
+      - name: Download JDK 17
+        run: |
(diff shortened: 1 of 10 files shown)


.github/workflows/build-pdb.yml
```diff
@@ -30,3 +30,51 @@ jobs:
       
     - name: Build with Maven
       run: mvn -B clean install --file pom.xml -DskipTests
+
+  build-jdk11:
+
+    runs-on: ubuntu-latest
```
TiagoMSSantos, line 36: I think we should force ubuntu 18 for now. WDYT?

.github/workflows/build-pdb.yml
```diff
@@ -30,3 +30,51 @@ jobs:
...
+
+      - name: Cache Maven packages
+        uses: actions/cache@v2.1.4
+        with:
+          path: ~/.m2
+          key: ${{ runner.os }}-m2-${{ hashFiles('**/pom.xml') }}-jdk11
+          restore-keys: ${{ runner.os }}-m2-jdk11
+
+      - name: Build with Maven
+        run: mvn -B clean install --file pom.xml -DskipTests
```
TiagoMSSantos, line 53: Can we use parallelism? `-T2C`

jcsf: This is a single module project. Parallelism doesn't make any difference.

.github/workflows/build-pdb.yml
jcsf: Why can't we just let this run on latest?

TiagoMSSantos: It was just a suggestion, because the latest can change any time by github when a new release of ubuntu is released.
But yeah, I guess it doesn't matter much.

TiagoMSSantos (approved):
lgtm

src/test/java/com/feedzai/commons/sql/abstraction/engine/impl/abs/SelectForUpdateTest.java
```diff
@@ -25,7 +25,6 @@
 import com.feedzai.commons.sql.abstraction.engine.impl.cockroach.SkipTestCockroachDB;
 import com.feedzai.commons.sql.abstraction.engine.testconfig.DatabaseConfiguration;
 import com.feedzai.commons.sql.abstraction.engine.testconfig.DatabaseTestUtil;
-import com.sun.org.apache.xpath.internal.operations.Bool;
```
artpdr, line 28: thanks XD

artpdr (approved):

PR #255, opened 2021-07-06
Disable the activation by default of Java 11 and 17 profiles

TiagoMSSantos:
This commit disables by default the activation of Java 11 and 17 profiles.
This is useful to avoid maven automatically activate Java 11 or 17 profile when the JDK detected is 11 or 17 and thus when compiling it produces an artifact that is not compatible with Java 8 anymore.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -786,9 +786,6 @@
         <!-- Profile to compile and run tests with JDK 11 -->
         <profile>
             <id>java-11</id>
-            <activation>
-                <jdk>11</jdk>
-            </activation>
             <properties>
                 <project.source>11</project.source>
             </properties>
@@ -808,9 +805,6 @@
         <!-- Profile to compile and run tests with JDK 17 -->
         <profile>
             <id>java-17</id>
-            <activation>
-                <jdk>17</jdk>
-            </activation>
             <properties>
                 <project.source>17</project.source>
             </properties>


hugomiguelabreu:
What does this means for our pipelines?

TiagoMSSantos:
> What does this means for our pipelines?

This is to make the deployment of PDB more safe, to avoid compiling it with the wrong Java version.
For our pipelines it doesn't matter much, since there we already control the Java version with the Github actions.

pom.xml
```diff
@@ -786,9 +786,6 @@
         <!-- Profile to compile and run tests with JDK 11 -->
         <profile>
             <id>java-11</id>
```
artpdr, line 788: Should we use <activeByDefault>false</activeByDefault> instead?

TiagoMSSantos: Its already `false` by default: https://maven.apache.org/ref/3.8.1/maven-settings/settings.html

artpdr (approved):

jcsf (changes requested):
Don't do this. With this, the pom's will not get changed in our pipelines that are running JDK 11 and 17. People must be carefull to not have java running on one of those versions when doing the release.

Which means that they are not compiling PDB in java 11/17 but instead in java 8 and then running it agains java 11/17.

TiagoMSSantos:
> Don't do this. With this, the pom's will not get changed in our pipelines that are running JDK 11 and 17. People must be carefull to not have java running on one of those versions when doing the release.
> 
> Which means that they are not compiling PDB in java 11/17 but instead in java 8 and then running it agains java 11/17.

The issue is that by compiling PDB with Java 11 will automatically activate Java 11 maven profile. Which means that the source and target versions will be Java 11, which then will make an artifact that is not compatible with Java 8 as it was intended (by not selecting any maven profile).
Without the `activation` tag, this will not happen in the future, since you can make an artifact compatible with Java 8 using Java 11 when compiling without any issue. This is just to prevent changing the source and target versions without activating a maven profile with intention for that.

jcsf:
> > Don't do this. With this, the pom's will not get changed in our pipelines that are running JDK 11 and 17. People must be carefull to not have java running on one of those versions when doing the release.
> > Which means that they are not compiling PDB in java 11/17 but instead in java 8 and then running it agains java 11/17.
> 
> The issue is that by compiling PDB with Java 11 will automatically activate Java 11 maven profile. Which means that the source and target versions will be Java 11, which then will make an artifact that is not compatible with Java 8 as it was intended (by not selecting any maven profile).
> Without the `activation` tag, this will not happen in the future, since you can make an artifact compatible with Java 8 using Java 11 when compiling without any issue. This is just to prevent changing the source and target versions without activating a maven profile with intention for that.

I understand that, but we want to ensure that we can move the code for JVM 11 so compiling it by updating the source and target is the best way to ensure that.

TiagoMSSantos:
> > > Don't do this. With this, the pom's will not get changed in our pipelines that are running JDK 11 and 17. People must be carefull to not have java running on one of those versions when doing the release.
> > > Which means that they are not compiling PDB in java 11/17 but instead in java 8 and then running it agains java 11/17.
> > 
> > 
> > The issue is that by compiling PDB with Java 11 will automatically activate Java 11 maven profile. Which means that the source and target versions will be Java 11, which then will make an artifact that is not compatible with Java 8 as it was intended (by not selecting any maven profile).
> > Without the `activation` tag, this will not happen in the future, since you can make an artifact compatible with Java 8 using Java 11 when compiling without any issue. This is just to prevent changing the source and target versions without activating a maven profile with intention for that.
> 
> I understand that, but we want to ensure that we can move the code for JVM 11 so compiling it by updating the source and target is the best way to ensure that.

When we officially support Java 11 by default, we can just exchange the default source and target versions by the ones in the Java 11 profile and change the profile name to Java 8.

jcsf:
> > > > Don't do this. With this, the pom's will not get changed in our pipelines that are running JDK 11 and 17. People must be carefull to not have java running on one of those versions when doing the release.
> > > > Which means that they are not compiling PDB in java 11/17 but instead in java 8 and then running it agains java 11/17.
> > > 
> > > 
> > > The issue is that by compiling PDB with Java 11 will automatically activate Java 11 maven profile. Which means that the source and target versions will be Java 11, which then will make an artifact that is not compatible with Java 8 as it was intended (by not selecting any maven profile).
> > > Without the `activation` tag, this will not happen in the future, since you can make an artifact compatible with Java 8 using Java 11 when compiling without any issue. This is just to prevent changing the source and target versions without activating a maven profile with intention for that.
> > 
> > 
> > I understand that, but we want to ensure that we can move the code for JVM 11 so compiling it by updating the source and target is the best way to ensure that.
> 
> When we officially support Java 11 by default, we can just exchange the default source and target versions by the ones in the Java 11 profile and change the profile name to Java 8.

The goal of this profiles are not the support but instead the testing. To ensure that until we upgrade java we don't break the compatibility with compiling and executing the PDB on Java 11/17

PR #272, opened 2021-10-11
Add support to Logger parameter when creating the batch

ctosin:
Change the batch to support a provided Logger object. If no Logger is provided, the default will still be used. A use case scenario for this is when a confidential logger is needed to prevent the batch to dump confidential information to outputs that are not allowed.

### DIFF
--- a/src/main/java/com/feedzai/commons/sql/abstraction/batch/AbstractBatch.java
+++ b/src/main/java/com/feedzai/commons/sql/abstraction/batch/AbstractBatch.java
@@ -437,9 +437,9 @@ public abstract class AbstractBatch implements Runnable {
         } catch (final Exception e) {
             if (this.maxFlushRetries > 0) {
                 final String msg = "[{}] Error occurred while flushing. Retrying.";
-                logger.warn(dev, msg, name);
+                confidentialLogger.warn(dev, msg, name, e);
                 if (confidentialLogger != logger) {
-                    confidentialLogger.warn(dev, msg, name, e);
+                    logger.warn(dev, msg, name);
                 }
             }
 
@@ -464,9 +464,9 @@ public abstract class AbstractBatch implements Runnable {
                     break;
                 } catch (final Exception ex) {
                     final String msg = "[{}] Error occurred while flushing (retry attempt {}).";
-                    logger.warn(dev, msg, name, retryCount + 1);
+                    confidentialLogger.warn(dev, msg, name, retryCount + 1, ex);
                     if (confidentialLogger != logger) {
-                        confidentialLogger.warn(dev, msg, name, retryCount + 1, ex);
+                        logger.warn(dev, msg, name, retryCount + 1);
                     }
                 }
             }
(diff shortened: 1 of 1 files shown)


jcsf (approved):

src/main/java/com/feedzai/commons/sql/abstraction/batch/AbstractBatch.java
```diff
@@ -169,7 +172,8 @@ protected AbstractBatch(
         final long maxAwaitTimeShutdown,
         @Nullable final BatchListener batchListener,
         final int maxFlushRetries,
-        final long flushRetryDelay) {
+        final long flushRetryDelay,
+        @Nullable final Logger logger) {
```
artpdr, line 176: Shouldn't this be a logger just for confidential logging purposes? We maintain the old logger to log non-confidential data and use the log that is passed here as an argument to log just confidential data... If this is received as null we log everything with the other logger that we had in the class before.

src/main/java/com/feedzai/commons/sql/abstraction/engine/DatabaseEngine.java
```diff
@@ -265,6 +266,23 @@
...
+     *
+     * @param batchSize     The batch size.
+     * @param batchTimeout  If inserts do not occur after the specified time, a flush will be performed.
+     * @param batchName     The batch name.
+     * @param batchListener Batch listener to execute custom behavior when the batch fails or succeeds to persist.
+     * @param logger        The logger.
+     * @return The batch.
+     */
+    default AbstractBatch createBatch(final int batchSize, final long batchTimeout, final String batchName, final BatchListener batchListener, final Logger logger) {
+        return createBatch(batchSize, batchTimeout, batchName, batchListener);
```
artpdr, line 283: this isn't passing the logger!?

src/main/java/com/feedzai/commons/sql/abstraction/batch/AbstractBatch.java
ctosin: The idea here is to allow a custom `Logger` to be sent as parameter without introducing any breaking changes to the current behavior PDB already has. Confidential logger is not a concept from PDB, it is from Pulse. That is why the code uses a generic Logger instance. After these changes are applied to PDB, Pulse will need to be changed as well, so it can call the correct `createBatch()`, which receives a confidential logger.

src/main/java/com/feedzai/commons/sql/abstraction/engine/DatabaseEngine.java
ctosin: I wrote the method this way because it is necessary to have a default implementation to avoid breaking changes to existing code. But I see your point, and this implementation seems "wrong", since it ignores the last parameter. After thinking more about that I decided to change the default implementation to throw an exception. That way we still have the default implementation and prevent method calls without explicitly override. That is the same approach Java's `Iterator.remove()` default method implementation uses.

artpdr (approved):

src/main/java/com/feedzai/commons/sql/abstraction/batch/AbstractBatch.java
```diff
@@ -399,7 +436,9 @@ public void flush() {
             logger.trace("[{}] Batch flushed. Took {} ms, {} rows.", name, (System.currentTimeMillis() - start), temp.size());
         } catch (final Exception e) {
             if (this.maxFlushRetries > 0) {
-                logger.warn(dev, "[{}] Error occurred while flushing. Retrying.", name, e);
+                final String msg = "[{}] Error occurred while flushing. Retrying.";
+                logger.warn(dev, msg, name);
+                confidentialLogger.warn(dev, msg, name, e);
```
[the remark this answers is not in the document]
ctosin: I added the checks. I was going to create a method to centralize that logic, but several logger methods are called with different types of parameters. So I decided to add a simple if block to prevent duplicate logging when a confidential logger is not provided.

src/main/java/com/feedzai/commons/sql/abstraction/engine/DatabaseEngine.java
```diff
@@ -265,6 +266,28 @@
...
+     * @param batchTimeout       If inserts do not occur after the specified time, a flush will be performed.
+     * @param batchName          The batch name.
+     * @param batchListener      Batch listener to execute custom behavior when the batch fails or succeeds to persist.
+     * @param confidentialLogger The confidential logger.
+     * @return The batch.
+     *
+     * @since 2.8.8
+     */
+    default AbstractBatch createBatch(final int batchSize, final long batchTimeout, final String batchName, final BatchListener batchListener, final Logger confidentialLogger) {
+        throw new UnsupportedOperationException("This method needs to be explicitly implemented ");
```
jcsf, line 288: I think you can by default just call a method that doesn't receive the logger:
```suggestion
        return createBatch(batchSize, batchTimeout, batchName, batchListener);
```

PR #273, opened 2021-10-20
Use version matrix in PDB CI actions

hugomiguelabreu:
In this commit a matrix version is included, ensuring more clean files and readable CI declarations.


hugomiguelabreu:
I have manually checked all the jobs, the java version is being correctly set and downloaded.
In the future we can also apply the same principle to the different databases, reducing the amount of action files to just 1 (one).

TiagoMSSantos (approved):
Nice one!

PR #287, opened 2021-12-15
AbstractDatabaseEngine function getConnection fixes

### LINKED ISSUE #286: AbstractDatabaseEngine.getConnection() - Debug log messages are incoherent and number of attempts is exceeded
In the function getConnection(), for a number of attempts equal to 10 and opening the log system for the debug mode it is noticed that the debug messages of attempt 5 and 9 are not presented (number of the middle and penultimate attempt), and in these attempts it only appears an error message that mentions that only X attempts are missing. Another behaviour noticed in this function, is that for a determined number of attempts, the function always executes one more attempt.

tmartinsfdz:
Fix debug messages and fix the number of attempts executed

Example of the new debug messages
image

Closes #286


tmartinsfdz:
Example of the new debug messages
image

PR #291, opened 2022-02-01
Bump H2 version and add support for H2v2

ctosin:
H2 v2.x is not compatible with v1.x. After the version bump, some changes to the code were necessary and the legacy mode was enabled. That increases compatibility between versions.


src/main/java/com/feedzai/commons/sql/abstraction/engine/impl/H2Engine.java
```diff
@@ -56,6 +58,10 @@
  * @since 2.0.0
  */
 public class H2Engine extends AbstractDatabaseEngine {
+    /**
+     * Tha max length of a VARCHAR type.
```
waldyrious, line 62:
```suggestion
     * The max length of a VARCHAR type.
```

src/main/java/com/feedzai/commons/sql/abstraction/engine/impl/H2Translator.java
```diff
@@ -355,4 +355,20 @@ public String translateTrue() {
     public String translateFalse() {
         return "FALSE";
     }
+
+    /**
+     * In H2 2.x, checking a boolean field against a boolean represented as a string causes an error.
+     * This code checks if the string is a boolean and process it as a boolean, not as a string.
```
waldyrious, line 361:
```suggestion
     * This code checks if the string is a boolean and processes it as a boolean, not as a string.
```

src/test/resources/connections.properties
```diff
@@ -5,14 +5,18 @@
 # the ports specified in the JDBC URLs are the default ones and are used by the Timeouts test
 ###############################################################################################
 
+# -=- H2 LEGACY MODE -=-
+# The version 2.x of H2 is not compatible with the version 1.x. When the version was bumped, the H2 legacy
+# mode was turned on to keep some of the backward compatibility with 1.x version. For more information, see:
```
waldyrious, line 10:
```suggestion
# Version 2.x of H2 is not compatible with version 1.x. When the version was bumped to 2.x,
# the H2 legacy mode was turned on to keep some backward compatibility with the 1.x version.
# For more information, see:
```

src/main/java/com/feedzai/commons/sql/abstraction/engine/impl/H2Engine.java
```diff
@@ -56,6 +58,10 @@
  * @since 2.0.0
  */
 public class H2Engine extends AbstractDatabaseEngine {
+    /**
+     * The max length of a VARCHAR type.
+     */
+    private static final int MAX_VARCHAR_LENGTH = 1048576;
```
francisco-polaco, line 64: Question: Why this value?

src/main/java/com/feedzai/commons/sql/abstraction/engine/impl/H2Translator.java
```diff
@@ -355,4 +355,20 @@ public String translateTrue() {
...
+
+    /**
+     * In H2 2.x, checking a boolean field against a boolean represented as a string causes an error.
+     * This code checks if the string is a boolean and processes it as a boolean, not as a string.
+     */
+    @Override
+    public String translate(final K k) {
+       if (k.getConstant() instanceof String) {
+            String s = (String) k.getConstant();
+            if (s.equalsIgnoreCase("TRUE") || s.equalsIgnoreCase("FALSE")) {
```
francisco-polaco, line 367: Question: Does `"1"` and `"0"` also represent boolean values?

src/main/java/com/feedzai/commons/sql/abstraction/engine/impl/H2Engine.java
ctosin: That value comes from the [documentation](http://www.h2database.com/html/datatypes.html#:~:text=CHARACTER%20VARYING,-%7B%20%7B%20CHARACTER%20%7C%20CHAR&text=create%20a%20quote.-,The%20allowed%20length%20is%20from%201%20to%201048576%20characters.,when%20using%20this%20data%20type.)

src/main/java/com/feedzai/commons/sql/abstraction/engine/impl/H2Translator.java
francisco-polaco: Makes sense. I was confused, because Oracle uses `1` and `0` to represent Booleans.

francisco-polaco (approved):
LGTM! :muscle: 

Please, add the link on the javadoc as suggested :)

src/main/java/com/feedzai/commons/sql/abstraction/engine/impl/H2Engine.java
francisco-polaco: Alright! I would add that link on the javadoc for reference :)

src/main/java/com/feedzai/commons/sql/abstraction/engine/impl/H2Translator.java
```diff
@@ -355,4 +355,20 @@ public String translateTrue() {
...
+                return Boolean.parseBoolean(s) ? translateTrue() : translateFalse();
+            }
+        }
```
jmf-tls, line 370: what if I have a string that happens to be "TRUE" but I really mean it as a string not a boolean??
instead of "TRUE", it's going to write in the DB "1"??

src/main/java/com/feedzai/commons/sql/abstraction/engine/impl/H2Engine.java
```diff
@@ -147,7 +153,16 @@ protected int entityToPreparedStatement(final DbEntity entity, final PreparedSta
 
                 switch (column.getDbColumnType()) {
                     case BLOB:
-                        ps.setBytes(i, objectToArray(val));
+                        if (val == null) {
+                            ps.setNull(i, Types.BLOB);
+                            break;
+                        }
+
+                        if (val instanceof Serializable) {
```
artpdr, line 162: I would use an `else if` here and remove the break from the block above...

src/test/resources/connections.properties
```diff
@@ -5,14 +5,18 @@
...
 ###############################################################################################
 
+# -=- H2 LEGACY MODE -=-
+# Version 2.x of H2 is not compatible with version 1.x. When the version was bumped to 2.x,
+# the H2 legacy mode was turned on to keep some backward compatibility with the 1.x version.
+# For more information, see: http://www.h2database.com/html/features.html.
 
 h2.engine=com.feedzai.commons.sql.abstraction.engine.impl.H2Engine
-h2.jdbc=jdbc:h2:./target/pdb;AUTO_SERVER=FALSE
+h2.jdbc=jdbc:h2:./target/pdb;AUTO_SERVER=FALSE;MODE=LEGACY
```
jmf-tls, line 14: This doesn't seem very bad for now, but in which ways is v2.x not compatible with v1.x?
Would it be too much trouble creating an implementation of H2Enginev2?

src/main/java/com/feedzai/commons/sql/abstraction/engine/impl/H2Engine.java
```diff
@@ -686,4 +701,20 @@ protected ResultIterator createResultIterator(PreparedStatement ps) throws Datab
     protected QueryExceptionHandler getQueryExceptionHandler() {
         return H2_QUERY_EXCEPTION_HANDLER;
     }
+
+    @Override
+    protected void setParameterValues(PreparedStatement ps, int index, Object param) throws SQLException {
```
artpdr, line 707: Can you make these final?

src/main/java/com/feedzai/commons/sql/abstraction/engine/impl/H2Engine.java
```diff
@@ -686,4 +701,20 @@ protected ResultIterator createResultIterator(PreparedStatement ps) throws Datab
...
+        if (param instanceof String) {
+            String paramStr = (String) param;
```
artpdr, line 709: final

src/main/java/com/feedzai/commons/sql/abstraction/engine/impl/H2Translator.java
```diff
@@ -355,4 +355,20 @@ public String translateTrue() {
...
     }
+
+    /**
+     * In H2 2.x, checking a boolean field against a boolean represented as a string causes an error.
+     * This code checks if the string is a boolean and process it as a boolean, not as a string.
+     */
+    @Override
+    public String translate(final K k) {
+       if (k.getConstant() instanceof String) {
+            String s = (String) k.getConstant();
```
artpdr, line 366: final

src/test/resources/connections.properties
```diff
@@ -5,14 +5,18 @@
...
 ###############################################################################################
 
+# -=- H2 LEGACY MODE -=-
+# The version 2.x of H2 is not compatible with the version 1.x. When the version was bumped, the H2 legacy
+# mode was turned on to keep some of the backward compatibility with 1.x version. For more information, see:
+# http://www.h2database.com/html/features.html.
 
 h2.engine=com.feedzai.commons.sql.abstraction.engine.impl.H2Engine
-h2.jdbc=jdbc:h2:./target/pdb;AUTO_SERVER=FALSE
+h2.jdbc=jdbc:h2:./target/pdb;AUTO_SERVER=FALSE;MODE=LEGACY
```
artpdr, line 14: Is it necessary for everyone using H2 on pdb to add this option?

artpdr:
On the commit message use just `Bump H2 version and use legacy mode for tests`

src/main/java/com/feedzai/commons/sql/abstraction/engine/impl/H2Engine.java
jmf-tls: I don't think we even need to have the special case for null here....
Check H2 driver internals, it seems to handle _null_ just fine

src/main/java/com/feedzai/commons/sql/abstraction/engine/impl/H2Translator.java
ctosin: The idea of this method is to consider "TRUE" or "FALSE" (strings) as TRUE or FALSE (booleans). Basically it removes the quotes. In a situation like this, the code would throw an error, because the unquoted value could not be stored in a string field.

I am aware that this implementation is not perfect. But CM is used to use booleans as strings in several select expressions, which is not allowed anymore in H2. So without this method, CM would have several breaking changes after the bump. So it is trade-off. In my opinion, it would be better to support that directly in PDB and from now on always use boolean comparisons as booleans, not strings.

src/test/resources/connections.properties
ctosin: The migration changes are documented here: https://www.h2database.com/html/migration-to-v2.html

I don't think that is necessary. I did tests with CM and Genome so far and none of them need the legacy mode. The legacy mode here in PDB is mainly because there are several tests that would fail without that (mainly because the way that H2 handled the ID counter in auto-increment columns, which changed in the new version). Fixing these tests is not trivial and that is why I opted to use the legacy mode.

src/test/resources/connections.properties
ctosin: No. The legacy mode is for testing purposes. PDB clients can choose to use it or not.

src/main/java/com/feedzai/commons/sql/abstraction/engine/impl/H2Engine.java
ctosin: The problem here is not the driver, but the `entry` map. This map may not have the key, so it is necessary to check for nulls.

src/main/java/com/feedzai/commons/sql/abstraction/engine/impl/H2Engine.java
ctosin: Yep!

src/main/java/com/feedzai/commons/sql/abstraction/engine/impl/H2Engine.java
ctosin: Yep!

src/main/java/com/feedzai/commons/sql/abstraction/engine/impl/H2Translator.java
ctosin: Yep!

src/main/java/com/feedzai/commons/sql/abstraction/engine/impl/H2Translator.java
ctosin: Ok, I see your point and I agree. I will remove that and clients of PDB will have to use boolean as boolean values.

src/test/resources/connections.properties
artpdr: I'm asking this because now we just test with legacy mode on. We aren't testing when the legacy mode is off which should happen in the majority of clients from this lib.

README.md
```diff
@@ -37,6 +37,11 @@ DatabaseEngine methods `#getPSResultSet(final String name)`, `#getPSIterator(fin
 (final String name, final int fetchSize)`, since 2.7.0, throw an extra `ConnectionResetException` when the database
 connection is lost and recovered.
 
+H2 2.x is not backward compatible with H2 1.x. Two changes worth some comments:
```
artpdr, line 40: I think this will require a release of PDB 2.9.0. So please put this text in a new header (## Changes from 2.9.0) after ## Changes from 2.8.0

ctosin: :+1:

src/test/java/com/feedzai/commons/sql/abstraction/engine/impl/h2/H2NotLegacyTest.java
```diff
@@ -0,0 +1,116 @@
...
+import static com.feedzai.commons.sql.abstraction.dml.dialect.SqlBuilder.dbEntity;
+import static com.feedzai.commons.sql.abstraction.dml.dialect.SqlBuilder.entry;
+import static com.feedzai.commons.sql.abstraction.util.StringUtils.quotize;
+import static org.junit.Assert.assertEquals;
+
+/**
+ * Tests H2 behavior in not legacy mode.
+ *
+ * @author Carlos Tosin (carlos.tosin@feedzai.com)
+ * @since ```feedzai.next.release```
```
artpdr, line 43: Don't use this in pdb. You will need to put the version hardcoded - 2.9.0

src/test/java/com/feedzai/commons/sql/abstraction/engine/impl/abs/EngineGeneralTest.java
```diff
@@ -168,7 +168,9 @@
 
     @Parameterized.Parameters
     public static Collection<DatabaseConfiguration> data() throws Exception {
-        return DatabaseTestUtil.loadConfigurations();
+        return DatabaseTestUtil.loadConfigurations(
+                "h2Legacy", "h2RemoteLegacy", "mysql", "sqlserver", "postgresql", "cockroach", "oracle", "db2"
```
artpdr, line 172: Why this change?

artpdr (commented):
I'm not comfortable with what we are doing with tests. We are testing just some things in the non legacy version. Some things that were tested before like in `BatchConnectionRetryTest` aren't being tested anymore.

src/test/java/com/feedzai/commons/sql/abstraction/engine/impl/abs/EngineGeneralTest.java
ctosin: Because this test class needs do load configurations of all of the database types, except non-legacy H2 (the specific non-legacy H2 tests are in another class - `H2NotLegacyTest`). So it is necessary to specify each one explicitly.

src/test/java/com/feedzai/commons/sql/abstraction/engine/impl/h2/H2NotLegacyTest.java
ctosin: 👍

.github/workflows/test-h2v2.yml
```diff
@@ -0,0 +1,36 @@
...
+on:
+  push:
+    branches: [ master ]
+  pull_request:
+    branches: [ master ]
+
+jobs:
+  test:
+    runs-on: ubuntu-latest
+    name: Test H22 PDB Java ${{ matrix.java-version }}
```
artpdr, line 15: H2v2

README.md
```diff
@@ -46,6 +46,11 @@ connection is lost and recovered.
 ## Changes from 2.8.0
 * It now uses Guava 25.1-jre version, which might require the client to also upgrade it to match the same version
 
+## Changes from 2.8.10
+* H2 version upgraded to 2.1.210
+* `H2Engine` is now deprecated and uses the H2 legacy mode (for more information see: http://www.h2database.com/html/features.html)
```
artpdr, line 51: ...the H2 v2 legacy mode if this engine is used with the h2 v2 driver...

src/main/java/com/feedzai/commons/sql/abstraction/engine/DatabaseEngineDriver.java
```diff
@@ -38,6 +38,20 @@ public String connectionStringFor(String hostname, String database, int port) {
...
+     * The H2v2 Database vendor.
+     */
+    H2V2("H2V2", "com.feedzai.commons.sql.abstraction.engine.impl.H2EngineV2", "org.h2.Driver", 0) {
+        @Override
+        public String connectionStringFor(String hostname, String database) {
+            return String.format("jdbc:h2:%s/%s", hostname, database);
+        }
+
+        @Override
+        public String connectionStringFor(String hostname, String database, int port) {
```
artpdr, line 51: We don't care about the port?

artpdr (approved):

README.md
ctosin: :+1:

src/main/java/com/feedzai/commons/sql/abstraction/engine/DatabaseEngineDriver.java
ctosin: I just followed the convention that was already being used by `H2`. I guess that's because this code relates to the embedded mode of H2, which does not use a port.

.github/workflows/test-h2v2.yml
ctosin: :+1:

jmf-tls (approved):

PR #297, opened 2022-02-14
Enable more tests

jmf-tls:
* Create missing profiles for H2, in pom and github workflows
* Add test classpath for H2 remote, so that UDF tests can run
* Make each H2 test config use different directories/ports
* Enable CockroachDB engine schema tests (required upgrading Docker image version, due to a bug https://github.com/cockroachdb/cockroach/issues/62601)
* Add flag to MySQL test configs to allow connecting to newer versions of the server
* Other improvements and cleanups

### DIFF
--- a/src/test/java/com/feedzai/commons/sql/abstraction/engine/testconfig/DatabaseTestUtil.java
+++ b/src/test/java/com/feedzai/commons/sql/abstraction/engine/testconfig/DatabaseTestUtil.java
@@ -20,6 +20,7 @@ import org.apache.commons.lang3.StringUtils;
 
 import java.io.FileInputStream;
 import java.io.InputStream;
+import java.util.Arrays;
 import java.util.Collection;
 import java.util.Set;
 import java.util.stream.Collectors;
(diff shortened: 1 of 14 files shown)


hugomiguelabreu:
Overall LGTM

ctosin (approved):

src/test/java/com/feedzai/commons/sql/abstraction/engine/testconfig/DatabaseTestUtil.java
```diff
@@ -66,25 +67,17 @@
...
-                    if (dbConfig.vendor.contains(vendor)) {
-                        return true;
-                    }
-                }
-
-                return false;
-            })
-            .collect(Collectors.toList());
+                .filter(dbConfig -> Arrays.stream(vendors).anyMatch(dbConfig.vendor::startsWith))
+                .collect(Collectors.toList());
```
upsidedownsmile, line 80: Nitty-gritty detail

This will be called several times(I think), why not load the configurations only once with a `Suppliers#memoize`?
Also you could use `Predicates.or` instead creating a stream to iterate over the array for each loaded configuration.

PR #299, opened 2022-02-15
Ensure queries respect fetch size

jmf-tls:
When querying the database with a fetch size set, some DB engines would ignore that setting and get all results into memory, which might crash the client code due to out of memory, or at least cause long GC pauses.

This commit makes all supported engines respect the fetch size:
* in the case of MySQL, server side cursors are enabled by setting the respective property in the connection configs;
* for PostgreSQL/CockroachDB the query is wrapped in a transaction if needed, so that a server side cursor is kept while the results are being fetched.


ctosin (approved):

src/main/java/com/feedzai/commons/sql/abstraction/dml/result/ResultIterator.java
```diff
@@ -82,14 +94,23 @@
...
             long start = System.currentTimeMillis();
+
+            this.isWrappedInTransaction = needsWrapInTransaction(statement.getFetchSize());
+
+            if (isWrappedInTransaction) {
+                final Connection connection = statement.getConnection();
+                this.previousAutocommit = connection.getAutoCommit();
+                connection.setAutoCommit(false);
+            } else {
+                this.previousAutocommit = false;
```
hugomiguelabreu, line 111: Why is this hard coded to `false`?

src/test/java/com/feedzai/commons/sql/abstraction/engine/impl/abs/TestRouter.java
```diff
@@ -0,0 +1,172 @@
...
+import static com.feedzai.commons.sql.abstraction.engine.configuration.PdbProperties.JDBC;
+
+/**
+ * A class that acts as a router, to forward data sent between a {@link DatabaseEngine} and the DB.
+ * <p>
+ * When running tests, this offers the possibility of interrupting connections without closing them.
+ *
+ * @author José Fidalgo (jose.fidalgo@feedzai.com)
+ */
+public class TestRouter implements Closeable {
```
artpdr, line 42: Cool

artpdr (approved):

src/main/java/com/feedzai/commons/sql/abstraction/dml/result/ResultIterator.java
```diff
@@ -56,20 +57,31 @@
     /**
      * The list of columns names for the given query.
      */
-    private final List<String> columnNames;
+    private final List<String> columnNames = new ArrayList<>();
```
upsidedownsmile, line 60: We don't need random access, so this could be a `LinkedList`.

src/main/java/com/feedzai/commons/sql/abstraction/dml/result/ResultIterator.java
```diff
@@ -333,4 +360,17 @@ private DatabaseEngineException closeAndHandleException(final Exception exceptio
     protected QueryExceptionHandler getQueryExceptionHandler() {
         return DEFAULT_QUERY_EXCEPTION_HANDLER;
     }
+
+    /**
+     * Indicates whether this iterator needs to run inside a transaction.
+     *
+     * PostgreSQL (and also CockroachDB) need this in order to keep a cursor on the server, otherwise the fetchsize
+     * setting is ignored and all results ar fetched from the database into memory at once.
```
upsidedownsmile, line 368: typo: `ar`

src/test/java/com/feedzai/commons/sql/abstraction/engine/impl/abs/TestRouter.java
upsidedownsmile, line 42: Why not make this `AutoCloseable`?

src/test/java/com/feedzai/commons/sql/abstraction/engine/impl/abs/TestRouter.java
```diff
@@ -0,0 +1,172 @@
...
+
+    /**
+     * Gets new DB properties with the JDBC URL patched to replace the default DB port with a new one.
+     *
+     * @param dbProperties The original DB properties.
+     * @param defaultPort  The default DB port.
+     * @param newPort      The port to use as replacement in the JDBC URL for the default DB port.
+     * @return The patched DB properties.
+     */
+    public static Properties getPatchedDbProperties(final Properties dbProperties, final int defaultPort, final int newPort) {
```
upsidedownsmile, line 94: This could be package-private.

src/test/java/com/feedzai/commons/sql/abstraction/engine/impl/abs/TestRouter.java
```diff
@@ -0,0 +1,172 @@
...
+        final Properties testProps = new Properties();
+        testProps.putAll(dbProperties);
+
+        final String patchedJdbc = dbProperties.getProperty(JDBC)
+                .replace("localhost:" + defaultPort, "localhost:" + newPort);
```
upsidedownsmile, line 99: This will just replace the port if the hostname is `localhost`. IMO either this just replace the port and change the javadocs  to say that it only replaces the port if the hostname is `localhost`

src/test/java/com/feedzai/commons/sql/abstraction/engine/impl/abs/TestRouter.java
```diff
@@ -0,0 +1,172 @@
...
+        executor.submit(() -> {
+                    final Socket socketTestToDb = socketFuture.get();
+                    logger.info("DB-to-Test established");
+                    while (isRunning) {
+                        int read = socketRouterToDb.getInputStream().read();
+                        socketTestToDb.getOutputStream().write(read);
+                    }
+                    return null;
+                }
+        );
```
upsidedownsmile, line 153: For this to work smoothly you would need an executor with at least 3 threads. But the executor is passed as an argument to the constructor, so someone could by mistake use an executor with only 1 or 2 threads.

IMO this executor should be internal to this class and instatiated in the constructor.

src/test/java/com/feedzai/commons/sql/abstraction/engine/impl/abs/TestRouter.java
upsidedownsmile, line 42: Wouldn't `TestProxy` be a more appropriate name?

src/test/java/com/feedzai/commons/sql/abstraction/engine/impl/abs/AbstractEngineSchemaTest.java
```diff
@@ -188,6 +194,59 @@ public void udfTimesTwoTest() throws Exception {
...
+                        .doesNotThrowAnyException();
+
+                // same as previous; ensures that the 3rd result is already fetched, and retries don't prevent from getting it
+                assertThatCode(iterator::next)
+                        .doesNotThrowAnyException();
+
+                // 4th row should be fetched now, but since the connection is broken it should fail
+                // (unless the driver already fetched all results, which should fail the test)
+                assertThatThrownBy(iterator::next)
+                        .isInstanceOf(Exception.class);
```
upsidedownsmile, line 242: Shouldn't this check the specific exception that is thrown when it can't be fetched from the DB?

upsidedownsmile (changes requested):

upsidedownsmile:
According to the MySql support team, 8.0 doesn't have this bug. So are we planning to not support 5.1 anymore?

src/main/java/com/feedzai/commons/sql/abstraction/dml/result/ResultIterator.java
jmf-tls: Yes we do; at least at line 239 we are using `columnNames.get(i)`.
I'd have to change that, and it would fall a bit outside of the scope of the issue being addressed

src/test/java/com/feedzai/commons/sql/abstraction/engine/impl/abs/TestRouter.java
jmf-tls: Closeable extends AutoCloseable, it is also auto-closed when used in try-with-resources. 
Usually I refer to the documentation, where it says for AutoCloseable:

> Note that unlike the {@link java.io.Closeable#close close} method of {@link java.io.Closeable}, this {@code close} method is <em>not</em> required to be idempotent.

And for Closeable:
> If the stream is already closed then invoking this method has no effect.

I want to explicitly state that the TestRouter#close behaves like the one in Closeable.

src/test/java/com/feedzai/commons/sql/abstraction/engine/impl/abs/TestRouter.java
jmf-tls: It could, but so could all the other public methods, or even the class itself.
Why are you picking on this specific method? :smile: 
Thing is, this TestRouter class used to be an inner class of TimeoutsTest and I had to extract it to make it available to AbstractEngineSchemaTest. This is just a test class, I don't want to restrict it like that and then require changes if some class in other package also needs to use it.

jmf-tls:
> According to the MySql support team, 8.0 doesn't have this bug. So are we planning to not support 5.1 anymore?

I checked, and it doesn't have the bug.
We are not currently using MySQL, but I we should keep support in PDB, I guess...
We could split this into a PDB API module and then implementation modules, that way we could have a mysql5 and a mysql8, each with a different driver version and different capabilities.
Or we could bump the driver version, adjust the code, and release a major version.

src/test/java/com/feedzai/commons/sql/abstraction/engine/impl/abs/TestRouter.java
upsidedownsmile: IMO method visibility should follow the same rules as security permissions, just enough for it to work. But this is a test class so I'm ok for it to be public.

(PS. I forgot to leave the same comment in the other methods :stuck_out_tongue: )

upsidedownsmile:
> > According to the MySql support team, 8.0 doesn't have this bug. So are we planning to not support 5.1 anymore?
> 
> I checked, and it doesn't have the bug. We are not currently using MySQL, but I we should keep support in PDB, I guess... We could split this into a PDB API module and then implementation modules, that way we could have a mysql5 and a mysql8, each with a different driver version and different capabilities. Or we could bump the driver version, adjust the code, and release a major version.

MySql 5.7 EOL will be next year(October 21, 2023) so maybe we could just support it until then and after just support 8.0

upsidedownsmile (approved):

PR #304, opened 2022-02-15
Fix EngineGeneralTest

jmf-tls:
In particular, the test `doesRowCountIncrementTest` was depending on data on a TEST table left over from a previous test. When adding a new test, the number of rows left over changed and the test started failing.
This commit makes the test create its own test data.


hugomiguelabreu (approved):

src/test/java/com/feedzai/commons/sql/abstraction/engine/impl/abs/EngineGeneralTest.java
```diff
@@ -1286,16 +1182,7 @@ public void maxTest() throws DatabaseEngineException {
 
     @Test
     public void minTest() throws DatabaseEngineException {
-        DbEntity entity = dbEntity()
```
artpdr, line 1289: Yes, it also hurts my eyes :stuck_out_tongue:

artpdr:
Arrived late but LGTM also...

PR #322, opened 2022-03-21
Make MetricsListener AutoCloseable

jmf-tls:
Make MetricsListener AutoCloseable so that the implementation knows when the batch is closed and can act upon it (e.g. unregister metrics).

### DIFF
--- a/src/main/java/com/feedzai/commons/sql/abstraction/batch/AbstractBatch.java
+++ b/src/main/java/com/feedzai/commons/sql/abstraction/batch/AbstractBatch.java
@@ -385,6 +385,7 @@ public abstract class AbstractBatch extends AbstractPdbBatch implements Runnable
     @Override
     public void close() throws Exception {
         destroy();
+        this.metricsListener.close();
     }
 
     /**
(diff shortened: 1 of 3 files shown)


src/main/java/com/feedzai/commons/sql/abstraction/batch/AbstractBatch.java
```diff
@@ -385,6 +385,7 @@ protected void start() {
     @Override
     public void close() throws Exception {
         destroy();
+        this.metricsListener.close();
```
ctosin, line 388: I am not sure if that is necessary, but consider if it would be necessary to check if `metricsListener` is not null.

src/main/java/com/feedzai/commons/sql/abstraction/batch/impl/MultithreadedBatch.java
```diff
@@ -241,6 +241,12 @@ public void close() {
         remainingTimeout = Math.max(this.maxAwaitTimeShutdownMs - (System.currentTimeMillis() - start), 1);
         orderlyShutdownExecutor(flusher, remainingTimeout);
 
+        try {
+            this.metricsListener.close();
```
ctosin, line 245: Same here.

ctosin (approved):

src/main/java/com/feedzai/commons/sql/abstraction/batch/AbstractBatch.java
jmf-tls: It shouldn't be.
I annotated the `BatchConfig.getMetricsListener()` with `@Nonnull`, meaning implementations should have something there; in the case of the `AbstractBatchConfig` with the current implementations extend, if the listener in the builder is null, then the final config will use `NoopMetricsListener`.
Same happens in the AbstractBatch deprecated constructors that have all separate parameters instead of BatchConfig.

src/main/java/com/feedzai/commons/sql/abstraction/batch/impl/MultithreadedBatch.java
jmf-tls: same as above

PR #382, opened 2023-04-11
Update dependency versions and do some cleanups

jmf-tls:
Summary:
- Updated some driver dependencies because of CVEs and to support newest database versions (still can't update MySQL driver, which also has CVE, because v8 isn't supported yet in PDB).
- Updated CockrachDB Docker image, with bugfixes and support for more SQL functions.
- Some tests were setting logger level to TRACE on BeforeClass, using Logback classes. Not only was this using Logback specific code (instead of SLF4J API), thus making updates more difficult, but the new log level would also affect other tests (since it is static). This commit removes that specific code and sets TRACE log level on the logback config file instead.

### DIFF
--- a/src/main/java/com/feedzai/commons/sql/abstraction/engine/impl/SqlServerEngine.java
+++ b/src/main/java/com/feedzai/commons/sql/abstraction/engine/impl/SqlServerEngine.java
@@ -669,7 +669,7 @@ public class SqlServerEngine extends AbstractDatabaseEngine {
             final String sString = format(
                     "SELECT T1.TABLE_NAME, CONSTRAINT_NAME "
                             + "FROM INFORMATION_SCHEMA.CONSTRAINT_TABLE_USAGE AS T1 "
-                            + "JOIN SYS.FOREIGN_KEYS AS F "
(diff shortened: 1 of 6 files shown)


src/main/java/com/feedzai/commons/sql/abstraction/engine/impl/SqlServerEngine.java
```diff
@@ -669,7 +669,7 @@ protected void dropReferringFks(DbEntity entity) throws DatabaseEngineException
             final String sString = format(
                     "SELECT T1.TABLE_NAME, CONSTRAINT_NAME "
                             + "FROM INFORMATION_SCHEMA.CONSTRAINT_TABLE_USAGE AS T1 "
-                            + "JOIN SYS.FOREIGN_KEYS AS F "
+                            + "JOIN sys.foreign_keys AS F "
```
jmf-tls, line 672: This doesn't change anything when the database is using a default collation that is case insensitive.  
However, when the default collation is case insensitive, only "sys.foreign_keys" works.  
In fact, JDBC has a method to get the foreign keys and we could use that, but would need some testing...

lmtjalves:
LGTM

Issue #384, opened 2023-05-23
It is not possible to persist NULL values to BLOB columns for Prepared Statements other than in H2


ruif95:
`AbstractDatabaseEngine#setPreparedStatementValue` is the method that different DB engines should override to set a given value onto a given index for a prepared statement. For each engine, it checks the data type and acts accordingly. For BLOB values, a byte stream is created from the value and the byte array is set as the PS (henceforth short for Prepared Statement) value for that position.

The problem occurs when this value is null. Only H2Engine has a safeguard in this regard, which checks if the value is null first and if so, sets the value to null in the PS. The other engines do not apply this safeguard, **meaning that the persisted value in the DB will not be null but instead be a small sequence of bytes**.

Each different DB engine impl should have this safeguard as well.

Example where this can affect tables:
- Prepared Statements that do not fill in anything on the BLOB column on purpose (for some given application logic): instead of having null values on the table, they will be filled in with a small value instead, all equal to each other since it's always the same null value byte array

PR #385, opened 2023-05-24
PDB-384: Make BLOB accept null values on all database engines

ruif95:
Summary:

Blob values, added to prepared statement value slots on a given index, are done so by creating a byte array stream of the value and passing the byte array to the value to be inserted. Without a null check, when persisting a null value what will actually end up on the column is a **small byte array representation of null**, and thus **the column will not actually be null** and hold a small value. The only current db system in which this does not happen is H2 because it already has a null check.

This commit adds a null check to each affected DB engine class and a general engine unit test with a scenario that goes through the affected codepath for any given engine type.

See more [here](https://github.com/feedzai/pdb/issues/384)

### DIFF
--- a/src/main/java/com/feedzai/commons/sql/abstraction/engine/impl/DB2Engine.java
+++ b/src/main/java/com/feedzai/commons/sql/abstraction/engine/impl/DB2Engine.java
@@ -56,6 +56,7 @@ import java.util.Set;
 import static com.feedzai.commons.sql.abstraction.util.StringUtils.md5;
 import static com.feedzai.commons.sql.abstraction.util.StringUtils.quotize;
 import static java.lang.String.format;
+import static java.util.Objects.isNull;
 import static org.apache.commons.lang3.StringUtils.join;
 
 /**
@@ -124,6 +125,11 @@ public class DB2Engine extends AbstractDatabaseEngine {
             case JSON:
             case CLOB:
             case BLOB:
+                if (isNull(value)) {
+                    ps.setBytes(index, null);
+                    break;
+                }
+
                 ps.setBytes(index, objectToArray(value));
 
                 break;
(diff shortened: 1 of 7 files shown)


src/test/java/com/feedzai/commons/sql/abstraction/engine/impl/abs/AbstractEngineSchemaTest.java
jmf-tls, line 1: You only added imports to this file, nothing else changed...  
Please revert

src/main/java/com/feedzai/commons/sql/abstraction/engine/impl/DB2Engine.java
```diff
@@ -124,6 +125,11 @@ protected void setPreparedStatementValue(final PreparedStatement ps,
             case JSON:
             case CLOB:
             case BLOB:
+                if (isNull(value)) {
+                    ps.setBytes(index, null);
+                    break;
+                }
```
jmf-tls, line 131: This looks ok to me, but since it's exactly the same in all implementations, it might as well be in AbstractDatabaseEngine.  
However, the same could be said about a lot more code, so...

src/test/java/com/feedzai/commons/sql/abstraction/engine/impl/abs/AbstractEngineSchemaTest.java
ruif95: Thanks for the notice! Reverted

jmf-tls (approved):
Previously this would store _null_ in a blob as a bunch of bytes, now it sends _null_, which brings some concerns regarding breaking changes.  
However, when reading the values, both cases will be correctly handled and a _null_ will be returned, so it should be ok.

joaonuno:
@ruif95 confirmed that reading from the previously stored format gives the same result, as long as the `toBlob()` method is used, as expected.

PR #386, opened 2023-05-24
Add Support to Upsert/Insert Ignore on PDB

victorcmg:
Add Support to Upsert/Insert Ignore on PDB

**Summary:**

- Introduces the capability to perform batch insert operations with duplicated entries without throwing a unique constraint violation exception. This is achieved by adding a new prepared statement to a `MappedEntity` when it is loaded. This statement will upsert the duplicate entries and will be implemented to each engine accordingly. This way, this prepared statement (upsert) will be used in the batch processing associated with the entity when adding and flushing the entries, but using the upserting any duplicated entries. This commit introduces support for the H2 and PostgreSQL engines for this operation.

- For the H2 engine, the prepared statement is the `MERGE` command which updates existing rows, and insert rows that don't exist. The documentation for this engine can be found [here](http://www.h2database.com/html/commands.html#merge_into)

```sql
MERGE INTO <TARGET_ENTITY> (cols...) VALUES(vals...)
```

- For the PostgreSQL engine, the insert statement introduces a `ON CONFLICT` clause that specifies an alternative action to raising a unique violation or exclusion constraint violation error. For the implementation, it was added the `DO UPDATE` action which will updating the columsn to the new values. It uses the values of the special table `EXCLUDED` to update it to the new values. The documentation for this engine can be found [here](https://www.postgresql.org/docs/current/sql-insert.html)

```sql
INSERT INTO <TARGET_ENTITY> (cols...) VALUES (vals...)
ON CONFLICT
DO UPDATE
SET (cols...) = EXCLUDED.(cols...)
```

**Test:**

- A new test was added to verify that a batch with duplicate entries successfully inserts only one entry and do not throw an exception.
- The test suite was run for every supported engine (H2 and PostgreSQL).

### DIFF
--- a/src/main/java/com/feedzai/commons/sql/abstraction/batch/AbstractBatch.java
+++ b/src/main/java/com/feedzai/commons/sql/abstraction/batch/AbstractBatch.java
@@ -372,8 +372,8 @@ public abstract class AbstractBatch extends AbstractPdbBatch implements Runnable
      * @param <R> the type of the second argument to the operation.
      */
     @FunctionalInterface
-    public interface ThrowingBiConsumer<T, R> {
-        void accept(T t, R r) throws Exception;
+    private interface FlushConsumer<T, R> {
+        void accept(T t, R r) throws DatabaseEngineException;
     }
 
     /**
@@ -480,7 +480,7 @@ public abstract class AbstractBatch extends AbstractPdbBatch implements Runnable
      *
      * @param processBatch A (throwing) BiConsumer to process the batch entries.
      */
-    private void flush(final ThrowingBiConsumer<DatabaseEngine, List<BatchEntry>> processBatch) {
+    private void flush(final FlushConsumer<DatabaseEngine, List<BatchEntry>> processBatch) {
         this.metricsListener.onFlushTriggered();
         final long flushTriggeredMs = System.currentTimeMillis();
         List<BatchEntry> temp;
@@ -539,7 +539,7 @@ public abstract class AbstractBatch extends AbstractPdbBatch implements Runnable
                         de.rollback();
                     }
 
-                    processBatch(de, temp);
+                    processBatch.accept(de, temp);
 
                     success = true;
                 } catch (final InterruptedException ex) {
--- a/src/main/java/com/feedzai/commons/sql/abstraction/batch/AbstractBatch.java
+++ b/src/main/java/com/feedzai/commons/sql/abstraction/batch/AbstractBatch.java
@@ -470,7 +470,7 @@ public abstract class AbstractBatch extends AbstractPdbBatch implements Runnable
     /**
      * Flushes the pending batches ignoring duplicate entries.
      */
-    public void flushIgnore() {
+    public void flushUpsert() {
         logger.trace("Start batch flushing ignoring duplicated entries.");
         flush((this::processBatchIgnoring));
     }
--- a/src/main/java/com/feedzai/commons/sql/abstraction/batch/AbstractBatch.java
+++ b/src/main/java/com/feedzai/commons/sql/abstraction/batch/AbstractBatch.java
@@ -15,22 +15,7 @@
  */
 package com.feedzai.commons.sql.abstraction.batch;
 
-import com.feedzai.commons.sql.abstraction.FailureListener;
-import com.feedzai.commons.sql.abstraction.engine.DatabaseEngine;
-import com.feedzai.commons.sql.abstraction.engine.DatabaseEngineException;
-import com.feedzai.commons.sql.abstraction.entry.EntityEntry;
-import com.feedzai.commons.sql.abstraction.listeners.BatchListener;
-import com.feedzai.commons.sql.abstraction.listeners.MetricsListener;
-import com.feedzai.commons.sql.abstraction.listeners.impl.NoopBatchListener;
-import com.feedzai.commons.sql.abstraction.listeners.impl.NoopMetricsListener;
 import com.google.common.base.Strings;
-import org.apache.commons.lang3.time.DurationFormatUtils;
-import org.slf4j.Logger;
-import org.slf4j.LoggerFactory;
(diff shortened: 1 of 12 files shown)


src/main/java/com/feedzai/commons/sql/abstraction/batch/AbstractBatch.java
```diff
@@ -365,6 +365,17 @@ protected AbstractBatch(final DatabaseEngine de, final int batchSize, final long
...
+    /**
+     * A functional interface to represent a {@link java.util.function.BiConsumer} that throws an exception.
+     *
+     * @param <T> the type of the first argument to the operation.
+     * @param <R> the type of the second argument to the operation.
+     */
+    @FunctionalInterface
+    public interface ThrowingBiConsumer<T, R> {
+        void accept(T t, R r) throws Exception;
+    }
```
diogo-anjos-fdz, line 377: If this is a public interface (even more a so generic one, both in the arguments as in the exception handling), then I would move this to a separate file.

If this is intended to be used just by this class, then I would 
- make it `private`,
- rename the method to something more use-case oriented, eg `FlushConsumer`,
- throw `DatabaseEngineException` instead

src/main/java/com/feedzai/commons/sql/abstraction/engine/impl/H2Engine.java
```diff
@@ -472,27 +473,54 @@ protected MappedEntity createPreparedStatementForInserts(final DbEntity entity)
...
+    /**
+     * Helper method to create a merge statement for this engine.
+     *
+     * @param entity    The entity.
+     * @param columns   The columns of this entity.
+     * @param values    The values of the entity.
+     *
+     * @return          A merge statement.
+     */
+    private String buildMergeStatement(final DbEntity entity, final List<String> columns, final List<String> values) {
```
diogo-anjos-fdz, line 507: I would rename to `buildUpsertStatement`.  
The fact this is being implemented through a `merge` statement is just an H2 under-the-hood implementation detail.

src/main/java/com/feedzai/commons/sql/abstraction/engine/impl/H2Engine.java
```diff
@@ -472,27 +473,54 @@ protected MappedEntity createPreparedStatementForInserts(final DbEntity entity)
...
+            psMerge = this.conn.prepareStatement(statementWithMerge, generateKeys);
             return new MappedEntity()
                 .setInsert(ps)
                 .setInsertReturning(psReturn)
                 .setInsertWithAutoInc(psWithAutoInc)
                 // The auto incremented column must be set, so when persisting a row, it's possible to retrieve its value
                 // by consulting the column name from this MappedEntity.
-                .setAutoIncColumn(columnWithAutoIncName);
+                .setAutoIncColumn(columnWithAutoIncName)
+                .setUpsert(psMerge);
```
diogo-anjos-fdz, line 492: See, if we rename the below method, we'll be avoiding this misalignment in the naming convention (ie. `.setUpsert(psUpsert)`).

src/main/java/com/feedzai/commons/sql/abstraction/batch/impl/MultithreadedBatch.java
```diff
@@ -340,6 +340,11 @@ When done, the future removes itself (if done already, all this can be skipped).
         }
     }
 
+    @Override
+    public void flushUpsert() {
+        logger.trace("Flush ignoring not available for MultithreadedBatch. Skipping ...");
+    }
```
diogo-anjos-fdz, line 346: This cannot be that dead silent :)

If this is not implemented and if it is still being used, then something is not right -- and then, the callee logic should _fail fast_, so it can be patched as fast as possible. So, I would throw an `UnsupportedOperationException`.

(at very least, this needs to be logged and `error`)

src/main/java/com/feedzai/commons/sql/abstraction/engine/impl/PostgreSqlEngine.java
```diff
@@ -404,23 +406,66 @@ protected MappedEntity createPreparedStatementForInserts(final DbEntity entity)
...
+    /**
+     * Helper method to create a insert on conflict statement that updates for this engine.
+     *
+     * @param entity    The entity.
+     * @param columns   The columns of this entity.
+     * @param values    The values of the entity.
+     *
+     * @return          A insert on conflict statement.
+     */
+    private String buildInsertOnConflictStatement(final DbEntity entity, final List<String> columns, final List<String> values) {
```
diogo-anjos-fdz, line 438: I would rename this to `..upsert..` as well.

src/main/java/com/feedzai/commons/sql/abstraction/batch/AbstractPdbBatch.java
```diff
@@ -72,4 +72,29 @@ protected void processBatch(final DatabaseEngine de, final List<BatchEntry> batc
...
+        de.beginTransaction();
+
+        for (final BatchEntry entry : batchEntries) {
+            de.addBatchUpsert(entry.getTableName(), entry.getEntityEntry());
+        }
+
+        de.flushUpsert();
+        de.commit();
+    }
+
```
diogo-anjos-fdz, line 99: DRY - this is duplicated code and should be improved.

(if there is time before the release, I would not make it a priority, just letting a note)

victorcmg: This requires some further analysis, right off the bat I can't change it as from what I saw it could take some refactoring.

diogo-anjos-fdz: ack! - we can address this later.

diogo-anjos-fdz (approved):
LGTM!

PR #388, opened 2023-06-22
Add Support to Upsert/Insert Ignore on PDB

victorcmg:
- Makes the implementation of DatabaseEngine optional, so it does not break compatibility with other library clients.


diogo-anjos-fdz (commented):
LGTM! Just left a minor comment.

diogo-anjos-fdz (approved):

PR #393, opened 2023-08-23
Add some logs while Postgres queries are being executed

TiagoMSSantos:
This commit adds some extra logs to be shown while Postgres queries are being executed. The goal is to easily identify cases where a query is not being executed because it's waiting for the Database table to be unlocked first (due to another query being already executed for example). Since it's not easy to identify a query timeout that will work for all cases of queries that can be executed with PDB, at least being able to see in the logs that PDB is still executing a query helps to identify such scenarios.

### DIFF
--- a/src/main/java/com/feedzai/commons/sql/abstraction/engine/impl/PostgreSqlEngine.java
+++ b/src/main/java/com/feedzai/commons/sql/abstraction/engine/impl/PostgreSqlEngine.java
@@ -37,7 +37,14 @@ import com.feedzai.commons.sql.abstraction.engine.handler.QueryExceptionHandler;
 import com.feedzai.commons.sql.abstraction.engine.impl.postgresql.PostgresSqlQueryExceptionHandler;
 import com.feedzai.commons.sql.abstraction.util.StringUtils;
 
+import java.time.Duration;
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Executors;
+import java.util.concurrent.ScheduledExecutorService;
+import java.util.concurrent.ScheduledFuture;
+import java.util.concurrent.TimeUnit;
 import java.util.stream.Collectors;
+
 import org.postgresql.Driver;
 import org.postgresql.PGProperty;
 import org.postgresql.util.PGobject;
@@ -112,6 +119,13 @@ public class PostgreSqlEngine extends AbstractDatabaseEngine {
      * @since 2.5.1
      */
     public static final QueryExceptionHandler PG_QUERY_EXCEPTION_HANDLER = new PostgresSqlQueryExceptionHandler();
+
+    /**
+     * The {@link ExecutorService} which will print some logs while Postgres {@link Statement update queries} are
+     * being executed. This will print a log from time to time in order to easily identify cases where the query might
+     * be stuck due to the database being locked because of another external query.
+     */
+    private final ScheduledExecutorService executor = Executors.newSingleThreadScheduledExecutor();
 
     /**
      * Creates a new PostgreSql connection.
@@ -236,24 +250,14 @@ public class PostgreSqlEngine extends AbstractDatabaseEngine {
 
         logger.trace(createTableStatement);
 
-        Statement s = null;
         try {
-            s = conn.createStatement();
-            s.executeUpdate(createTableStatement);
+            executeUpdateQuery(createTableStatement);
         } catch (final SQLException ex) {
             if (ex.getSQLState().startsWith(NAME_ALREADY_EXISTS)) {
                 logger.debug(dev, "'{}' is already defined", entity.getName());
                 handleOperation(new OperationFault(entity.getName(), OperationFault.Type.TABLE_ALREADY_EXISTS), ex);
             } else {
                 throw new DatabaseEngineException("Something went wrong handling statement", ex);
-            }
-        } finally {
-            try {
-                if (s != null) {
-                    s.close();
-                }
-            } catch (final Exception e) {
-                logger.trace("Error closing statement.", e);
             }
         }
     }
@@ -283,24 +287,14 @@ public class PostgreSqlEngine extends AbstractDatabaseEngine {
 
         logger.trace(addPrimaryKey);
 
-        Statement s = null;
         try {
-            s = conn.createStatement();
-            s.executeUpdate(addPrimaryKey);
+            executeUpdateQuery(addPrimaryKey);
         } catch (final SQLException ex) {
             if (ex.getSQLState().startsWith(TABLE_CAN_ONLY_HAVE_ONE_PRIMARY_KEY)) {
                 logger.debug(dev, "'{}' already has a primary key", entity.getName());
                 handleOperation(new OperationFault(entity.getName(), OperationFault.Type.PRIMARY_KEY_ALREADY_EXISTS), ex);
             } else {
                 throw new DatabaseEngineException("Something went wrong handling statement", ex);
-            }
-        } finally {
-            try {
-                if (s != null) {
-                    s.close();
-                }
-            } catch (final Exception e) {
-                logger.trace("Error closing statement.", e);
             }
         }
     }
@@ -334,24 +328,14 @@ public class PostgreSqlEngine extends AbstractDatabaseEngine {
 
             logger.trace(statement);
 
-            Statement s = null;
             try {
-                s = conn.createStatement();
-                s.executeUpdate(statement);
+                executeUpdateQuery(statement);
             } catch (final SQLException ex) {
                 if (ex.getSQLState().startsWith(NAME_ALREADY_EXISTS)) {
                     logger.debug(dev, "'{}' is already defined", idxName);
                     handleOperation(new OperationFault(entity.getName(), OperationFault.Type.INDEX_ALREADY_EXISTS), ex);
                 } else {
                     throw new DatabaseEngineException("Something went wrong handling statement", ex);
-                }
-            } finally {
-                try {
-                    if (s != null) {
-                        s.close();
-                    }
-                } catch (final Exception e) {
-                    logger.trace("Error closing statement.", e);
                 }
             }
         }
@@ -495,12 +479,10 @@ public class PostgreSqlEngine extends AbstractDatabaseEngine {
 
     @Override
     protected void dropTable(DbEntity entity) throws DatabaseEngineException {
-        Statement drop = null;
+        final String query = format("DROP TABLE %s CASCADE", quotize(entity.getName()));
+        logger.trace(query);
         try {
-            drop = conn.createStatement();
-            final String query = format("DROP TABLE %s CASCADE", quotize(entity.getName()));
-            logger.trace(query);
-            drop.executeUpdate(query);
+            executeUpdateQuery(query);
         } catch (final SQLException ex) {
             if (ex.getSQLState().startsWith(TABLE_OR_VIEW_DOES_NOT_EXIST)) {
                 logger.debug(dev, "Table '{}' does not exist", entity.getName());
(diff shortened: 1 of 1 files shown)


src/test/java/com/feedzai/commons/sql/abstraction/engine/impl/abs/EngineIsolationTest.java
```diff
@@ -562,6 +562,11 @@ protected DatabaseEngine[] prepareEngineForDeadlockRecoveryTest() throws Excepti
         engines[0] = DatabaseFactory.getConnection(properties);
         final EntityEntry entry = entry().set("COL1", 1).set("COL2", 0).build();
 
+        if (this.engineDriver.equals(DatabaseEngineDriver.ORACLE)) {
+            // Fix ORA-00604: error occurred at recursive SQL level 1ORA-08177: can't serialize access for this transaction
+            engines[0].getConnection().setTransactionIsolation(Connection.TRANSACTION_READ_COMMITTED);
```
TiagoMSSantos, line 567: This is just to fix some Oracle tests that sometimes fail due to the error:
```
ORA-00604: error occurred at recursive SQL level 1ORA-08177: can't serialize access for this transaction
``

**Review by helenapoleri — APPROVED** [review]

File: src/test/java/com/feedzai/commons/sql/abstraction/engine/impl/abs/EngineIsolationTest.java (hunk 1 above)
Comment (artpdr) [reply src/test/java/com/feedzai/commons/sql/abstraction/engine/impl/abs/EngineIsolationTest.java:567]: Well this method is setting the transaction level as serializable which is the highest isolation level. What's the reasoning to set the transaction level for one of the connections to READ COMMITED?

File: src/main/java/com/feedzai/commons/sql/abstraction/engine/impl/PostgreSqlEngine.java [hunk 2]
```diff
@@ -532,11 +546,12 @@ protected void dropColumn(DbEntity entity, String... columns) throws DatabaseEng
         }
         removeColumns.add(join(cols, ","));
 
+        final AtomicReference<Statement> queryStatement = new AtomicReference<>();
```
artpdr, line 549: Why do we need the atomic references here and in the other lines?

src/main/java/com/feedzai/commons/sql/abstraction/engine/impl/PostgreSqlEngine.java
```diff
@@ -832,4 +847,38 @@ protected ResultIterator createResultIterator(PreparedStatement ps) throws Datab
...
+        final long startMilliSeconds = System.currentTimeMillis();
+        final Runnable printLog = () -> {
+            final long secondsPassed = (System.currentTimeMillis() - startMilliSeconds) / 1000;
+            this.logger.info("Database is still executing for '{}s' a query.\n" +
+                "Are you sure the Database is not blocked due to another query being executed?",
+                secondsPassed
+            );
+            this.logger.trace("The query still being executed for '{}s' is: '{}'.\n", secondsPassed, queryLog);
+        };
+        final ScheduledFuture<?> scheduledFutureLog = this.executor.scheduleAtFixedRate(printLog, 5L, 5L, TimeUnit.MINUTES);
```
artpdr, line 872: What happens when this is called concurrently or when a task is already scheduled to run?

src/main/java/com/feedzai/commons/sql/abstraction/engine/impl/PostgreSqlEngine.java
```diff
@@ -832,4 +847,38 @@ protected ResultIterator createResultIterator(PreparedStatement ps) throws Datab
...
+     *
+     * @param function      The {@link Callable function} that will executes Postgres {@link Statement query}.
+     * @param queryLog      The query in {@link String} format which is being executed. Will be used in the logs.
+     * @throws SQLException If the {@link Statement query} fails during its execution.
+     * @implNote It prints a log every 5 minutes so the user is aware that the query is still being executed.
+     */
+    private<T> T executeUpdateQuery(final Callable<T> function, final String queryLog) throws SQLException {
+        final long startMilliSeconds = System.currentTimeMillis();
+        final Runnable printLog = () -> {
+            final long secondsPassed = (System.currentTimeMillis() - startMilliSeconds) / 1000;
```
artpdr, line 865: nitpick: You could also use `Duration` or something similar.

TiagoMSSantos: Done! :)

src/main/java/com/feedzai/commons/sql/abstraction/engine/impl/PostgreSqlEngine.java
TiagoMSSantos: Ok, removed! Now the new method `com.feedzai.commons.sql.abstraction.engine.impl.PostgreSqlEngine#executeUpdateQuery` closes the `Statement`.

helenapoleri (approved):

src/test/java/com/feedzai/commons/sql/abstraction/engine/impl/abs/EngineIsolationTest.java
artpdr: But why this fixes it? Why oracle can't have the two connections serialized?

src/main/java/com/feedzai/commons/sql/abstraction/engine/impl/PostgreSqlEngine.java
artpdr: Where do you see it as synchronous? Even if it has any lock inside that makes it synchronous it is probably only synchronous on new updates done with the same statement. However, we do a `this.conn.createStatement()` that creates a new Statement object per connection. Am I missing something?

src/test/java/com/feedzai/commons/sql/abstraction/engine/impl/abs/EngineIsolationTest.java
TiagoMSSantos: To be honest, I'm not sure about this. Do you think we should revert this?

TiagoMSSantos: Reverted.

src/main/java/com/feedzai/commons/sql/abstraction/engine/impl/PostgreSqlEngine.java
TiagoMSSantos: Well the new method `com.feedzai.commons.sql.abstraction.engine.impl.PostgreSqlEngine#executeUpdateQuery` will only return when the `Statement` is executed so that makes it synchronous. WDYT?

src/main/java/com/feedzai/commons/sql/abstraction/engine/impl/PostgreSqlEngine.java
```diff
@@ -582,20 +546,10 @@ protected void addColumn(DbEntity entity, DbColumn... columns) throws DatabaseEn
         final String addColumnsStatement = join(addColumns, " ");
         logger.trace(addColumnsStatement);
 
-        Statement s = null;
         try {
-            s = conn.createStatement();
-            s.executeUpdate(addColumnsStatement);
-        } catch (final SQLException ex) {
+            executeUpdateQuery(addColumnsStatement);
+        } catch (final Throwable ex) {
```
artpdr, line 551: SQLException

src/main/java/com/feedzai/commons/sql/abstraction/engine/impl/PostgreSqlEngine.java
```diff
@@ -495,34 +480,22 @@ protected void dropSequences(DbEntity entity) throws DatabaseEngineException {
 
     @Override
     protected void dropTable(DbEntity entity) throws DatabaseEngineException {
-        Statement drop = null;
         try {
-            drop = conn.createStatement();
             final String query = format("DROP TABLE %s CASCADE", quotize(entity.getName()));
             logger.trace(query);
```
artpdr: mov eout of try please

src/main/java/com/feedzai/commons/sql/abstraction/engine/impl/PostgreSqlEngine.java
```diff
@@ -533,24 +506,15 @@ protected void dropColumn(DbEntity entity, String... columns) throws DatabaseEng
         removeColumns.add(join(cols, ","));
 
         try {
-            drop = conn.createStatement();
             final String query = join(removeColumns, " ");
             logger.trace(query);
```
artpdr: move out of try please

src/main/java/com/feedzai/commons/sql/abstraction/engine/impl/PostgreSqlEngine.java
```diff
@@ -678,11 +632,9 @@ protected void addFks(final DbEntity entity, Set<DbFk> fks) throws DatabaseEngin
                             quotizedForeignColumnsString
                     );
 
-            Statement alterTableStmt = null;
             try {
-                alterTableStmt = conn.createStatement();
                 logger.trace(alterTable);
```
artpdr: move out of try please

src/main/java/com/feedzai/commons/sql/abstraction/engine/impl/PostgreSqlEngine.java
```diff
@@ -832,4 +776,46 @@ protected ResultIterator createResultIterator(PreparedStatement ps) throws Datab
...
+     * executed and the current one is waiting for that one to finish first.
+     *
+     * @param query The query in {@link String} format which will be executed.
+     * @throws SQLException If the {@link Statement query} fails during its execution.
+     * @implNote It prints a log every 5 minutes so the user is aware that the query is still being executed.
+     */
+    private int executeUpdateQuery(final String query) throws SQLException {
+        final long startMilliSeconds = System.currentTimeMillis();
+        final Runnable printLog = () -> {
+            final Duration timePassed = Duration.of(System.currentTimeMillis() - startMilliSeconds, ChronoUnit.MILLIS);
```
artpdr, line 793: I would change this to 
`long timePassed = Duration.ofMillis(System.currentTimeMillis()).minusMillis(startMilliSeconds).toMinutes();`

and then use it directly in the logs below.

src/main/java/com/feedzai/commons/sql/abstraction/engine/impl/PostgreSqlEngine.java
TiagoMSSantos: Done!

src/main/java/com/feedzai/commons/sql/abstraction/engine/impl/PostgreSqlEngine.java
TiagoMSSantos: Done!

src/main/java/com/feedzai/commons/sql/abstraction/engine/impl/PostgreSqlEngine.java
TiagoMSSantos: Done!

src/main/java/com/feedzai/commons/sql/abstraction/engine/impl/PostgreSqlEngine.java
TiagoMSSantos: Done!

src/main/java/com/feedzai/commons/sql/abstraction/engine/impl/PostgreSqlEngine.java
```diff
@@ -832,4 +775,46 @@ protected ResultIterator createResultIterator(PreparedStatement ps) throws Datab
...
+        final ScheduledFuture<?> scheduledFutureLog = this.executor.scheduleAtFixedRate(printLog, 5L, 5L, TimeUnit.MINUTES);
+
+        Statement statement = null;
+        try {
+            statement = this.conn.createStatement();
+            return statement.executeUpdate(query);
+        } catch (final SQLException ex) {
+            throw ex;
+        } catch (final Exception ex) {
+            throw new RuntimeException(ex);
```
artpdr, line 808: Remove these catches please...

TiagoMSSantos: Done!

artpdr (approved):

PR #407, opened 2023-10-13
Removes error logging for UPSERT command when creating an entity

victorcmg:
This commit removes an unnecessary stack trace logging when an entity is created on pdb and it has no primary keys.

### DIFF
--- a/src/main/java/com/feedzai/commons/sql/abstraction/engine/impl/DB2Engine.java
+++ b/src/main/java/com/feedzai/commons/sql/abstraction/engine/impl/DB2Engine.java
@@ -444,7 +444,7 @@ public class DB2Engine extends AbstractDatabaseEngine {
                         .setUpsert(psUpsert);
 
         } catch (final IllegalArgumentException e) {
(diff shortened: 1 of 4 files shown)


diogo-anjos-fdz:
LGTM!

diogo-anjos-fdz (approved):

victorcmg-fdz:
Just so it keeps documented, the tests sometimes are failing on the H2 engine. 

The same behavior could be seen locally and it seems related to the latest automatic bumps on the `h2` dependency. The tests made rolled-back (locally) to the previous version before the bump, e.g. `2.1.214` and all tests succeeded. On the other hand, when bumping the dependency to its latest version `2.2.224` the tests tend to fail. The same behavior was seen during the tests in this PR and also throughout the commit history since the bumps have been committed to the main branch. So, most likely, the flakiness regarding the pdb master branch is due to the bumping of this library (and even maybe other ones).

The test ran locally were a way to sanity check that this PR would not be introducing any other issue to the main branch.

PR #471, opened 2024-08-20
Reverts the Support to Upsert/Insert Ignore on PDB

victorcmg-fdz:
Recently, it was added a support for the _upsert of entries_ in the `pdb` library. 

The feature added supported for entities to, when applicable, be able to be _upserted_ or depending on the engine implementation, an _insert ignore_ or even _merge_. The feature also enabled the upsert feature for batch processing, i.e. entries could now use the new upsert API. However, after its usage it was possible to conclude it decreased performance when client application used this new API. Giving the degrading performance and also the log messages some of the entities generated for not having the upsert statement available, it was decided to **remove** this **feature** entirely from the `pdb` code-base.

This series of commit reverts the feature to upsert entries on `pdb`. The reverted commits reverted are as follows.

- 94ba1a5
- 46f29b2
- 963cd30
- d4f7c4e
- d626297
- 36a434c
- 1658965
- d2f153b
- 2f0d1ff
- 348a22d
- e100a51

### DIFF
--- a/.github/workflows/test-oracle.yml
+++ b/.github/workflows/test-oracle.yml
@@ -15,7 +15,6 @@ jobs:
     name: Test Oracle PDB Java ${{ matrix.java-version }}
 
     strategy:
-      max-parallel: 1
       matrix:
         java-version: [ 8, 11, 17 ]
     steps:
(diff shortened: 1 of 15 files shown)


diogo-anjos-fdz:
LGTM!

Just one quick note: please improve the PR summary with the rational/business need to revert these changes.  
That is, are we reverting these because we are following a new direction, are we preparing for a refactoring, are we deprecating the feature?

victorcmg-fdz:
> LGTM!
> 
> Just one quick note: please improve the PR summary with the rational/business need to revert these changes. That is, are we reverting these because we are following a new direction, are we preparing for a refactoring, are we deprecating the feature?

good call! updated!

PR #527, opened 2026-06-22
fix: Only print warn/error logs in confidentialLogger if its a different logger

TiagoMSSantos:
This change is avoid print PII sensitive data from the exceptions into the standard logger.

### DIFF
--- a/src/main/java/com/feedzai/commons/sql/abstraction/batch/impl/MultithreadedBatch.java
+++ b/src/main/java/com/feedzai/commons/sql/abstraction/batch/impl/MultithreadedBatch.java
@@ -364,9 +364,9 @@ public class MultithreadedBatch extends AbstractPdbBatch implements PdbBatch {
         } catch (final Exception e) {
             if (this.maxFlushRetries > 0) {
                 final String msg = "[{}] Error occurred while flushing. Retrying.";
-                confidentialLogger.warn(DEV, msg, name, e);
+                logger.warn(DEV, msg, name);
                 if (confidentialLogger != logger) {
-                    logger.warn(DEV, msg, name);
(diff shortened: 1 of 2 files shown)


samyuh (approved):

src/main/java/com/feedzai/commons/sql/abstraction/batch/impl/MultithreadedBatch.java
```diff
@@ -364,9 +364,9 @@ private void flush(final long flushTriggeredMs, final List<BatchEntry> batchEntr
         } catch (final Exception e) {
             if (this.maxFlushRetries > 0) {
                 final String msg = "[{}] Error occurred while flushing. Retrying.";
-                confidentialLogger.warn(DEV, msg, name, e);
+                logger.warn(DEV, msg, name);
```
artpdr, line 367: I don’t  think this and other log levels must be changed at all which was done in a previous PR. I think they should continue as trace as they may contain PII.

TiagoMSSantos: The PII sensitive data will only be logged for confidential logger. Can you elaborate why change this to error / warn is an issue for confidential logger?

TiagoMSSantos: Asking because we only have INFO as default log level, so without this change, we would still have to enable trace logs to know when some DB query fails :thinking:

TiagoMSSantos: Or do you mean 'name' can also contain PII? and it should be trace because of that?

